PR #8, opened 2019-01-08
Support + testing jacobian and adjoint computations

willtebbutt:
What is says on the tin.

### DIFF
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -3,7 +3,7 @@
 
 Approximate the gradient of `f` at `x` using `fdm`. Assumes that `f(x)` is scalar.
 """
-function grad(fdm, f, x::AbstractArray{T}) where T
+function grad(fdm, f, x::AbstractArray{T}) where T<:Real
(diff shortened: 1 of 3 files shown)


willtebbutt:
I'll fix up the coverage before you review @wesselb

willtebbutt:
I've added an additional test to my new code, but I have literally no idea why the coverage has dropped for the `numerics.jl` file. Any thoughts @ararslan @wesselb ?

ararslan:
There are coverage emission issues with Julia itself in 0.7 and 1.0 that have been addressed in 1.1. Currently (and unfortunately), massive, unexplainable changes in coverage are to be expected at the moment.

willtebbutt:
@wesselb any thoughts on this?

willtebbutt:
Following on from our earlier slack conversation, I think that the name `adjoint` really _does_ make sense here -- it is the adjoint operator of the forwards-mode operator (also known as the `primal` operator I believe). Since the forwards-mode operator is a linear (differential) operator, the adjoint is too. Does this make sense, or am I chatting nonsense?

PR #9, opened 2019-01-25
Proposal: Tighten types

willtebbutt:
`similar` is only guaranteed to do sensible things for `Arrays`, as opposed to any arbitrary `AbstractArray`. Consequently, I propose constrained the types of the arrays involved in `gradient` / `jacobian` / `jvp` / `j′vp` to reflect this.

This problem actually caught me out when working with `BlockArrays`. `similar(block_vector)` produces a `BlockVector` whose indices are undefined because the memory for the data isn't allocated, which caused some undefined reference issues.

@wesselb @ararslan what are your thoughts on this?

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,0 +1,15 @@
+# This file is machine-generated - editing it directly is not advised
+
+[[Libdl]]
+uuid = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+
+[[LinearAlgebra]]
+deps = ["Libdl"]
+uuid = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+
+[[Printf]]
+deps = ["Unicode"]
+uuid = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+
+[[Unicode]]
+uuid = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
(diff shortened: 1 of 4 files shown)


src/grad.jl
```diff
@@ -3,7 +3,7 @@
 
 Approximate the gradient of `f` at `x` using `fdm`. Assumes that `f(x)` is scalar.
 """
-function grad(fdm, f, x::AbstractArray{T}) where T<:Real
+function grad(fdm, f, x::Array{T}) where T<:Real
```
wesselb: Maybe turn `Array` into `Vector` to be consistent with the below?
```suggestion
function grad(fdm, f, x::Vector{T}) where T<:Real
```

Manifest.toml
```diff
@@ -0,0 +1,15 @@
+# This file is machine-generated - editing it directly is not advised
```
ararslan: Generally, Manifest.toml files for packages should not be checked into Git but rather listed in .gitignore.

ararslan:
> whose indices are undefined

I'm not sure what you mean. Do you mean that the values produced by iterating `eachindex` are sometimes out of bounds? If so, that sounds like an issue with that particular package.

ararslan:
Hm, kind of odd but okay. IMO I think it makes more sense for the user to be mindful of the limitations of the particular array type they're using rather than us requiring it to be a standard `Array`. A case I assume would be not particularly uncommon is `SubArray`s, which result from `view`.

PR #10, opened 2019-01-26
 Simplify the method order check to avoid a try/catch 

ararslan:
In general it's best not to rely on errors for control flow. For example, it could be the case in some future version that `factorial` could wrap around instead of error on overflow. Further, some `AbstractVector` could return a `BigInt` from `length`, in which case this error would never occur; you'd just get really, really big numbers.

This also adds tests for the error conditions in `fdm` and removes the Manifest file added in #9.


wesselb:
Also looks good to me. Thanks for the comments!

PR #11, opened 2019-04-13
Implements extensions to jvp and adjoint

willtebbutt:
Relates to [this PR](https://github.com/JuliaDiff/ChainRules.jl/pull/14). Makes it easier to compute jacobian-vector product and jacobian-transpose-vector products.

### DIFF
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -35,15 +35,81 @@ end
 jacobian(fdm, f, x::Vector{<:Real}) = jacobian(fdm, f, x, length(f(x)))
 
 """
-    jvp(fdm, f, x::AbstractVector{<:Real}, ẋ::AbstractVector{<:Real})
+    jvp(fdm, f, x::Vector{<:Real}, ẋ::AbstractVector{<:Real})
 
 Convenience function to compute `jacobian(f, x) * ẋ`.
 """
-jvp(fdm, f, x::Vector{<:Real}, ẋ::AV{<:Real}) = jacobian(fdm, f, x) * ẋ
+_jvp(fdm, f, x::Vector{<:Real}, ẋ::AV{<:Real}) = jacobian(fdm, f, x) * ẋ
 
 """
-    j′vp(fdm, f, ȳ::AbstractVector{<:Real}, x::AbstractVector{<:Real})
+    j′vp(fdm, f, ȳ::AbstractVector{<:Real}, x::Vector{<:Real})
 
 Convenience function to compute `jacobian(f, x)' * ȳ`.
 """
-j′vp(fdm, f, ȳ::AV{<:Real}, x::Vector{<:Real}) = jacobian(fdm, f, x, length(ȳ))' * ȳ
+_j′vp(fdm, f, ȳ::AV{<:Real}, x::Vector{<:Real}) = jacobian(fdm, f, x, length(ȳ))' * ȳ
+
+
+
+
+# Transform `x` into a vector, and return a closure which inverts the transformation.
+to_vec(x::Real) = ([x], x->x[1])
+
+# Arrays.
+to_vec(x::Vector{<:Real}) = (x, identity)
+to_vec(x::Array) = vec(x), x_vec->reshape(x_vec, size(x))
+
+# AbstractArrays.
+function to_vec(x::T) where {T<:LinearAlgebra.AbstractTriangular}
+    x_vec, back = to_vec(Matrix(x))
+    return x_vec, x_vec->T(reshape(back(x_vec), size(x)))
+end
+to_vec(x::Symmetric) = vec(Matrix(x)), x_vec->Symmetric(reshape(x_vec, size(x)))
+to_vec(X::Diagonal) = vec(Matrix(X)), x_vec->Diagonal(reshape(x_vec, size(X)...))
+
+# Non-array data structures.
+function to_vec(x::Tuple)
+    x_vecs, x_backs = zip(map(to_vec, x)...)
+    sz = cumsum([map(length, x_vecs)...])
+    return vcat(x_vecs...), function(v)
+        return ([x_backs[n](v[sz[n]-length(x[n])+1:sz[n]]) for n in 1:length(x)]...,)
+    end
+end
+
+"""
+    jvp(fdm, f, x, ẋ)
+
+Compute a Jacobian-vector product with any types of arguments for which `to_vec` is defined.
+"""
+function jvp(fdm, f, x, ẋ)
+    x_vec, vec_to_x = to_vec(x)
+    ẋ_vec, _ = to_vec(ẋ)
+    y = f(x)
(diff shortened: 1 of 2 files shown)


src/grad.jl
```diff
@@ -35,15 +35,66 @@ end
...
+end
+to_vec(x::Symmetric) = vec(Matrix(x)), x_vec->Symmetric(reshape(x_vec, size(x)))
+to_vec(X::Diagonal) = vec(Matrix(X)), x_vec->Diagonal(reshape(x_vec, size(X)...))
+
+# Non-array data structures.
+function to_vec(x::Tuple)
+    x_vecs, x_backs = zip(map(to_vec, x)...)
+    sz = cumsum([map(length, x_vecs)...])
+    return vcat(x_vecs...), function(v)
+        return ([x_backs[n](v[sz[n]-length(x[n])+1:sz[n]]) for n in 1:length(x)]...,)
```
ararslan:
```suggestion
        return ntuple(n->x_backs[n](v[sz[n]-length(x[n])+1:sz[n]]), length(x))
```
I haven't done any rigorous benchmarking of this or anything, but it should at least avoid materializing an intermediate array.

test/grad.jl
```diff
@@ -1,29 +1,83 @@
...
-@testset "jacobian / jvp / j′vp" begin
-    rng, P, Q, fdm = MersenneTwister(123456), 3, 2, central_fdm(5, 1)
-    ȳ, A, x, ẋ = randn(rng, P), randn(rng, P, Q), randn(rng, Q), randn(rng, Q)
-    Ac = copy(A)
+    @testset "jvp" begin
+        rng, N, M, fdm = MersenneTwister(123456), 2, 3, central_fdm(5, 1)
+        x, y = randn(rng, N), randn(rng, M)
+        ẋ, ẏ = randn(rng, N), randn(rng, M)
+        xy, ẋẏ = vcat(x, y), vcat(ẋ, ẏ)
+        ż_manual = _jvp(fdm, (xy)->sum(sin.(xy)), xy, ẋẏ)[1]
```
ararslan:
```suggestion
        ż_manual = _jvp(fdm, (xy)->sum(sin, xy), xy, ẋẏ)[1]
```
Same applies elsewhere. It's just a bit more efficient as it avoids making two passes over the input, one to broadcast and one to reduce.

ararslan: (Not that it really matters here, as these are just tests...)

src/grad.jl
```diff
@@ -35,15 +35,66 @@ end
...
+"""
+function j′vp(fdm, f, ȳ, x)
+    x_vec, vec_to_x = to_vec(x)
+    ȳ_vec, _ = to_vec(ȳ)
+    return vec_to_x(_j′vp(fdm, x_vec->to_vec(f(vec_to_x(x_vec)))[1], ȳ_vec, x_vec))
+end
+j′vp(fdm, f, ȳ, xs...) = j′vp(fdm, xs->f(xs...), ȳ, xs)
+
+# Transform `x` into a vector, and return a closure which inverts the transformation.
+to_vec(x::Real) = ([x], x->x[1])
```
ararslan:
```suggestion
to_vec(x::Real) = ([x], first)
```
Not a big deal or anything, just seems nice to be able to avoid the anonymous function.

src/grad.jl
```diff
@@ -35,15 +35,66 @@ end
...
+    x_vec, back = to_vec(Matrix(x))
+    return x_vec, x_vec->T(reshape(back(x_vec), size(x)))
+end
+to_vec(x::Symmetric) = vec(Matrix(x)), x_vec->Symmetric(reshape(x_vec, size(x)))
+to_vec(X::Diagonal) = vec(Matrix(X)), x_vec->Diagonal(reshape(x_vec, size(X)...))
+
+# Non-array data structures.
+function to_vec(x::Tuple)
+    x_vecs, x_backs = zip(map(to_vec, x)...)
+    sz = cumsum([map(length, x_vecs)...])
```
ararslan: Unfortunate that there isn't a more convenient way to express this with `cumsum`, which is defined only on arrays and does not accept a function argument like `sum`. Perhaps an issue to Julia itself would be worthwhile.

src/grad.jl
```diff
@@ -35,15 +35,66 @@ end
...
+end
+function jvp(fdm, f, xẋs::Tuple{Any, Any}...)
+    x, ẋ = collect(zip(xẋs...))
+    return jvp(fdm, xs->f(xs...), (x, ẋ))
+end
+
+"""
+    j′vp(fdm, f, ȳ, x...)
+
+Compute an adjoint with any types of arguments for which `to_vec` is defined.
```
ararslan: It would probably be worthwhile to have a docstring for `to_vec` since it's not exported and is otherwise not discoverable.

src/grad.jl
```diff
@@ -35,15 +35,66 @@ end
...
+
+Compute a Jacobian-vector product with any types of arguments for which `to_vec` is defined.
+"""
+function jvp(fdm, f, (x, ẋ)::Tuple{Any, Any})
+    x_vec, vec_to_x = to_vec(x)
+    _, vec_to_y = to_vec(f(x))
+    return vec_to_y(_jvp(fdm, x_vec->to_vec(f(vec_to_x(x_vec)))[1], x_vec, to_vec(ẋ)[1]))
+end
+function jvp(fdm, f, xẋs::Tuple{Any, Any}...)
+    x, ẋ = collect(zip(xẋs...))
```
ararslan: I'm having a difficult time following this part of the code. Note that in the function signature, you're restricting `xẋs` to be a 2-tuple and splatting it, which means that the signature is just
```julia
function jvp(fdm, f, x, ẋ)
```
Now, the line below the signature is equivalent to
```julia
x, ẋ = collect(zip(x, ẋ))
```
If `x` or `ẋ` has more than 2 elements, this is discarding the rest, which means this is equivalent to
```julia
x, ẋ = (x[1], ẋ[1]), (x[2], ẋ[2])
```
Then so what's getting passed to the next call is a tuple of 2-tuples.

Is that analysis correct and the behavior is intentional, or have I missed something here?

willtebbutt: Hmm possibly I've messed up here. This signature is meant to be shorthand for
```julia
function jvp(fdm, f, (x, xdot), (y, ydot), (z, zdot), ...)
```
so that each `Tuple{Any, Any}` contains an argument-differential pair. As a result, the (un)zipping operation should produce
```julia
x, ẋ = (x, y, z), (xdot, ydot, zdot)
```

willtebbutt: Ah, scratch that, got the wrong thing to mind. Your analysis is correct. Not that I've also wrapped `f` in an anonymous function that splats out the arguments. This makes it possible to handle varargs functions by converting them into unary functions of `Tuple`s.

PR #13, opened 2019-04-18
Add to_vec for Adjoint and Transpose

ararslan:
Currently `to_vec` fails for inputs wrapped in these types, so this adds a method that handles them.


src/grad.jl
```diff
@@ -94,6 +94,11 @@ end
 to_vec(x::Symmetric) = vec(Matrix(x)), x_vec->Symmetric(reshape(x_vec, size(x)))
 to_vec(X::Diagonal) = vec(Matrix(X)), x_vec->Diagonal(reshape(x_vec, size(X)...))
 
+function to_vec(X::T) where T<:Union{Adjoint,Transpose}
+    U = T.name.wrapper
```
ararslan: Coverage drop is likely because this is being optimized out.

PR #14, opened 2019-04-21
to_vec Tuple bug (do not merge)

willtebbutt:
The length of the wrong object was being used. This is fine most of the time, but not when the length of `x` and `x_vec` are different. Will find a decent test to add to prevent reintroduction of the bug.


willtebbutt:
There is a bug in the Tuple code, but my changes to `Diagonal` are totally wrong. Will resolve, please don't merge.

willtebbutt:
Superceded by #20

PR #17, opened 2019-05-16
Add adaptive function bound computation

ararslan:
This introduces a type-based approach that works slightly differently than the current API. Upon construction of a `FDMethod` subtype, the order of the method and the order of the derivative are checked for agreement, and the grid and coefficients are computed and stored. Instances of the types are callable, and doing so produces the value of the derivative. It works recursively when adaptively computing the bound, which results in a better step size and thus a (hopefully) more robust estimate of the derivative.

"Why types?" you may be asking. It just seemed like an efficient way of expressing things and storing intermediate values.

"Where did this adaptive thing come from?" From none other than Wessel himself: https://github.com/wesselb/fdm.

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -1,4 +1,4 @@
-export FDMReport, fdm, backward_fdm, forward_fdm, central_fdm
+export fdm, backward_fdm, forward_fdm, central_fdm
 
 """
     FDM.DEFAULT_CONDITION
@@ -26,6 +26,11 @@ function central_grid(p::Int)
     end
 end
 
+"""
+    History
+
+A mutable type that tracks several values during adaptive bound computation.
+"""
 mutable struct History
     adapt::Int
     eps::Real
@@ -42,10 +47,41 @@ mutable struct History
     end
 end
 
+"""
+    FDMethod
+
+Abstract type for all finite differencing method types.
+Subtypes of `FDMethod` are callable with the signature
+
+```
+method(f, x; kwargs...)
+```
+
+where the keyword arguments can be any of
+
+* `adapt`: The number of adaptive steps to use improve the estimate of `bound`.
+* `bound`: Bound on the value of the function and its derivatives at `x`.
+* `condition`: The condition number. See [`DEFAULT_CONDITION`](@ref).
+* `eps`: The assumed roundoff error. Defaults to `eps()` plus [`TINY`](@ref).
+"""
 abstract type FDMethod end
 
-for D in (:Forward, :Backward, :Central)
-    gridf = Symbol(lowercase(String(D)), "_grid")
+function Base.show(io::IO, x::FDMethod)
+    @printf io "FDMethod:\n"
+    @printf io "  order of method:       %d\n" x.p
+    @printf io "  order of derivative:   %d\n" x.q
+    @printf io "  grid:                  %s\n" x.grid
+    @printf io "  coefficients:          %s\n" x.coefs
+    h = x.history
+    if all(p->isdefined(h, p), propertynames(h))
+        @printf io "  roundoff error:        %.2e\n" h.eps
+        @printf io "  bounds on derivatives: %.2e\n" h.bound
+        @printf io "  step size:             %.2e\n" h.step
+        @printf io "  accuracy:              %.2e\n" h.accuracy
+    end
+end
+
+for D in (:Forward, :Backward, :Central, :Nonstandard)
     @eval begin
         struct $D{G<:AbstractVector, C<:AbstractVector} <: FDMethod
             p::Int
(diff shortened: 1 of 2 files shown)


ararslan:
I've marked this as a draft because I'd like feedback in general, but especially on:
* The implementation. Is this the best way to go about computing the bound?
* The API. Should this replace the current API? How could it be improved? Currently it duplicates all existing logic and lives separately from the existing API.
* Should we rename and reuse `FDMReport` for this? We could make it mutable and have it update the stored values on each adaptation step.
* What are some corner cases that could be used to test that this provides improvements over the non-adaptive approach?

src/methods.jl
```diff
@@ -1,5 +1,88 @@
...
+    adapt=1,
+    report=false,
+) where M<:FDMethod
+    p = m.p
+    q = m.q
+    grid = m.grid
+    coefs = m.coefs
+
+    # Adaptively compute the bound on the function and derivative values, if applicable.
+    if 0 < adapt < p
```
wesselb: What is the motivation for the upper bound in this check?

src/methods.jl
```diff
@@ -1,5 +1,88 @@
...
+        newm = (M.name.wrapper)(p + 1, q)
```
wesselb: I think `bound` should be set to an upper bound on the `order`th (i.e. `p`th) derivative, so I believe this should be `newm = (M.name.wrapper)(p + 1, p)`. Here `p + 1` comes from that the order must be strictly greater than the derivative to estimate. Furthermore, `order` in our FDM setup refers to the number of grid points, which is easily interpretable, but which is also one off from what it typically means. Perhaps we should adjust this.

src/methods.jl
```diff
@@ -1,5 +1,88 @@
...
+        dfdx, = fdm(newm, f, x; eps=eps, bound=bound, adapt=(adapt - 1))
+        bound = maxabs(dfdx)
```
wesselb: We have to be careful here to ensure that `bound > 0`. A simple solution is to add a tiny number, e.g. `1e-20`. (Note that the case `bound = 0` can easily occur, e.g. when adapting a `p=2` and `q=1` method for a linear function.)

src/methods.jl
```diff
@@ -1,5 +1,88 @@
...
+end
+
+maxabs(x) = maximum(abs, x)
+
+function fdm(
+    m::M,
+    f,
+    x;
+    bound=maxabs(f(x)),
+    eps=Base.eps(bound),
```
wesselb: We also have to be careful here that `eps > 0`. Again the simple solution is to add a tiny number, e.g. `1e-20`.

wesselb:
First of all, this looks great!!

> The implementation. Is this the best way to go about computing the bound?

I have added some comments regarding the implementation.

It might also be worthwhile to implement the [condition number mechanism](https://github.com/wesselb/fdm/blob/master/fdm/fdm.py#L22) to make the procedure even more robust. I think robustness of the estimation is an important feature.

> The API. Should this replace the current API? How could it be improved? Currently it duplicates all existing logic and lives separately from the existing API.

I think the new API with the types is pretty sensible and could replace the current API. We might still want to support the old API by redirecting to the new API to maintain compatibility.

> Should we rename and reuse `FDMReport` for this? We could make it mutable and have it update the stored values on each adaptation step.

Yeah, I'm not sure if having a separate type (i.e. `FDMReport`) just to store this information is the best way to go about it. Perhaps the new types (i.e. `Forward` etc) can instead store this information, which would be overwritten at every computation, thus always referring to the last computation. Any ideas on this?

> What are some corner cases that could be used to test that this provides improvements over the non-adaptive approach?

A nice example is the estimation of the derivative of `x -> 1 / x` at an `x` near zero:

```python
>>> central_fdm(5, 1, adapt=0)(lambda x: 1 / x, 1e-3) + 1e6
670283.9150891255

>>> central_fdm(5, 1, adapt=1)(lambda x: 1 / x, 1e-3) + 1e6
2.133892849087715e-05

>>> central_fdm(5, 1, adapt=2)(lambda x: 1 / x, 1e-3) + 1e6
5.097826942801476e-07
```

In general, I have not found much use of having more than two adaptation steps, and `adapt=1` should probably be the default. For difficult functions, having a high number of adaptation steps may even prove harmful, because that requires estimating high-order derivatives, which can fail catastrophically for difficult functions.

Furthermore, the cases `bound = 0` and `eps = 0` should be tested. Having `bound > 0` and `eps > 0` is absolutely critical. For example,

```python
>>> central_fdm(3, 2, adapt=0)(lambda x: x, 1.0)
0.0
```

gives `0.0` *exactly*, and this could introduce division by zero, if we're not careful.

src/methods.jl
ararslan: We require that `p` be less than 20 so that `factorial(p)` doesn't error (since `p` is a regular `Int`, not a `BigInt`), and inside of this condition we're recursing with `p` set to `p + 1`. So we can only do `p - adapt` adaptation steps before it would error.

wesselb: If every adaptation step increases `p` by one, then, for `p <= 20` to hold, we can do at most `20 - p` adaptation steps, right? Also, it might be more sensible to throw an error or warning if `adapt` is too big; currently, in that case, the method would just silently perform no adaptation, which might not be expected.

ararslan: > we can do at most `20 - p` adaptation steps, right?

Ah yeah, whoops. :grimacing: 

> Also, it might be more sensible to throw an error or warning if `adapt` is too big

Agreed :+1:

src/methods.jl
ararslan: `Base.eps` will always be nonzero, so I think in this case it should be sufficient to simply add a check that `eps > 0` and throw an error otherwise.

wesselb: Ah, that is true. I'm not sure if `Base.eps(0.0) == 5.0e-324` is too small though. We should check that.

ararslan: Too small in what sense? It's small enough that e.g. `1 / eps(0.0)` is `Inf`, which I suppose is problematic.

wesselb: I'm worried about the below computation on `h\hat`. In general, working near the edges of the floating point ranges is dangerous. For example `Base.eps(0.0) / 10 == 0.0`, so stuff could go very wrong if `C2` is large. Adaptation could make this even more complicated.

ararslan: Heh, yeah:
```
julia> Central(5, 1)(sin, 1; adapt=1, eps=eps(0.0))
-Inf
```

wesselb:
This looks really good to me now! Let's see what others have to say.

Another example where adaptation is useful is estimating the derivative of `log` near `0`:

```python
>>> forward_fdm(5, 1, adapt=0)(np.log, 1e-3)
996.0418094849194

>>> forward_fdm(5, 1, adapt=1)(np.log, 1e-3)
999.9999999445137
```

src/methods.jl
```diff
@@ -1,5 +1,128 @@
...
+derivatives.
+"""
+const DEFAULT_CONDITION = 100
+
+"""
+    FDM.TINY
+
+A tiny number added to some quantities to ensure that division by 0 does not occur.
+"""
+const TINY = 1e-20
```
willtebbutt: Maybe we want to replace this constant by `eps()`? Or have I misunderstood what this is going to be used for?

src/methods.jl
```diff
@@ -1,5 +1,128 @@
...
+        return vcat(div(-p, 2):-1, 1:div(p, 2))
+    end
+end
+
+mutable struct History
+    adapt::Int
+    eps::Real
+    bound::Real
+    step::Real
+    accuracy::Real
```
willtebbutt: Do we care that the types of `eps`, `bound`, etc are abstract? Does this likely have performance implications?

ararslan: It does, but this is a mutable type anyway, so it will already not be terribly performant. The problem with having them concrete is that when the `History` object is constructed, we don't know the types of the values it will end up holding, since this is constructed before we call the `FDMethod` on a function. The function could return `Float32`, for example, and we wouldn't necessarily want to promote to `Float64`. (Though a case could be made to do so.)

src/methods.jl
ararslan: Yeah, we could do that. I was just following what's done in the Python implementation.

wesselb: The tiny number here should be absolutely small, so I'm not sure `eps()` makes sense. What we had before, `eps(0.0)`, would be ideal, if it not were at the edge of the floating point range. `1e-20` is a compromise to have something absolutely small to what we usually work with, whilst not it being too small to cause issues with the numerics.

wesselb: Perhaps the above comments doesn't really make sense. `TINY` is supposed to be small compared to any number that we typically encounter, but `eps()` is only small to numbers of the order of magnitude of `1`. If we consider `1e-6` on the lower end of sizes of numbers we typically encounter, then `eps(1e-6) = 2e-22` should be sufficient, which is where I originally got the `1e-20` from. Does that make more sense? 😬

ararslan:
Another thing I haven't yet done here is have the current API simply call the new API; the code for both exists and is largely identical.

ararslan:
> For example,
>
> ```python
> >>> central_fdm(3, 2, adapt=0)(lambda x: x, 1.0)
> 0.0
> ```
> gives `0.0` _exactly_, and this could introduce division by zero, if we're not careful.

```julia
julia> FDM.Central(3, 2, adapt=0)(identity, 1.0)
0.0

julia> central_fdm(3, 2)(identity, 1.0)
7.277897151650305e-7
```
Hmmmmm.

wesselb:
Hmm, interesting. The Python implementation gives me a different result:

```python
>>> central_fdm(3, 2, adapt=0)(lambda x: x, 1.0)
0.0

>>> central_fdm(3, 2, adapt=1)(lambda x: x, 1.0)
0.0
```

I'll have a closer look at the code.

ararslan:
The behavior implemented here matches Python, but what's odd is that the original implementation does not.

wesselb:
Oh, I totally missed that the second call uses the old API; didn't have my morning coffee yet...

I think it's due to the condition number:

```julia
julia> central_fdm(3, 2; M=1)(x -> x, 1.0)
7.277897151650305e-7

julia> central_fdm(3, 2; M=100)(x -> x, 1.0)
0.0
```

ararslan:
Okay, assuming CI agrees, I think things should be in good shape now. Unfortunately method handling was almost entirely rewritten. :neutral_face: Also note that I've added a `Nonstandard` method which represents a custom grid, otherwise we would lose that functionality with this change, however we can't do adaptive bounds computation on a custom grid.

ararslan:
UGH. Adaptively computing the bounds actually results in _more_ test failures in ChainRules. I was hoping it would make it more robust to changes in the random seed.

ararslan:
This no longer destroys Nabla's tests once very minor adjustments are made in Nabla to facilitate this change.

wesselb:
This looks very good to me! Happy to merge, if everyone else is so as well.

PR #26, opened 2019-07-05
add multi arg jacobian

Roger-luo:
XRef: https://github.com/FluxML/Zygote.jl/pull/235

I'm not sure what kind of test should be added here?


oxinabox:
@willtebbutt  can you review?

willtebbutt:
Definitely needs some tests.

> I'm not sure what kind of test should be added here?

I would compare the results of the multi-arg `jacobian` with multiple calls to the single arg jacobian. i.e. something like
```julia
foo(x, y) = # some binary function
binary_result = jacobian(fdm, foo, x, y)
x_result = jacobian(fdm, x->foo(x, y), x)
y_result = jacobian(fdm, y->foo(x, y), y)

# some code to compare x_result and y_result with binary_result
```

willtebbutt:
@Roger-luo sorry for leaving this without attention for so long. Are you planning to continue working on it, or have you moved on to other things?

Roger-luo:
Oh, sry. I didn't see your reply. Yeah I'll revise this. One sec.

Roger-luo:
the original interface of `jacobian` is changed due to conflict with varargs, and I make the `grad` function more generic (not just `Vector`) but multi args as well.

Roger-luo:
@willtebbutt seems good to go. let me know if I need to change anything.

Roger-luo:
BTW, I think my implementation of `grad` can handle all `to_vec` now. Do you want me to just make use of it? It seems `to_vec` is some sort of workaround on the old `grad` which can only take `Vector` as argument. The new `grad` should just work for a bunch of things without defining anything else.

src/grad.jl
```diff
@@ -1,40 +1,64 @@
 export grad, jacobian, jvp, j′vp, to_vec
+function replace_arg(x, xs::Tuple, k::Int)
```
nickrobinson251: style: blank line before function

 ```suggestion

function replace_arg(x, xs::Tuple, k::Int)
```

**Comment by willtebbutt** [comment]
> BTW, I think my implementation of grad can handle all to_vec cases. now. Do you want me to just make use of it? It seems to_vec is some sort of workaround on the old grad which can only take Vector as argument. The new grad should just work for a bunch of things without defining anything else.

I think `to_vec` is still important here. For example, have you tried your new implementation of `grad` with a `Dict`? My gut tells me that you'll need to use `to_vec` to successfully handle everything. I would be interested to know if this isn't the case for some reason though.

File: src/grad.jl [hunk 2]
```diff
@@ -1,40 +1,64 @@
...
+    return ntuple(length(xs)) do p
+        if p == k
+            x
+        else
+            xs[p]
+        end
+    end
+end
```
nickrobinson251: why not put on one line as `ntuple(p -> p == k ? x : xs[p], length(xs))` ?

src/grad.jl
```diff
@@ -1,40 +1,64 @@
...
-                tmp .= x .+ ϵ .* v
-                return f(tmp)
-            end,
-            zero(T),
-        )
-        v[n] = zero(T)
+function grad end
+
+function grad(fdm, f, x::AbstractArray{T}) where T
+    dx, tmp = similar(x), similar(x)
```
nickrobinson251: style: clearer on two lines
```suggestion
    dx = similar(x)
    tmp = similar(x)
```

src/grad.jl
```diff
@@ -1,40 +1,64 @@
...
+function grad(fdm, f, xs...)
+    return ntuple(length(xs)) do k
+        grad(fdm, x->f(replace_arg(x, xs, k)...), xs[k])
+    end
+end
+
 """
-    jacobian(fdm, f, x::Vector{<:Number}, D::Int)
-    jacobian(fdm, f, x::Vector{<:Number})
+    jacobian(fdm, f, xs::Union{Real, AbstractArray{<:Real}}[; dim::Int=length(f(x))])
```
nickrobinson251: style: given the keyword has a default, the "optional" syntax just adds noise

```suggestion
    jacobian(fdm, f, xs::Union{Real, AbstractArray{<:Real}}; dim::Int=length(f(x)))
```

src/grad.jl
```diff
@@ -1,40 +1,64 @@
...
 
 Approximate the Jacobian of `f` at `x` using `fdm`. `f(x)` must be a length `D` vector. If
 `D` is not provided, then `f(x)` is computed once to determine the output size.
 """
-function jacobian(fdm, f, x::Vector{T}, D::Int) where {T<:Number}
-    J = Matrix{T}(undef, D, length(x))
-    for d in 1:D
-        J[d, :] = grad(fdm, x->f(x)[d], x)
+function jacobian(fdm, f, x::Union{T, AbstractArray{T}}; dim::Int=length(f(x))) where {T <: Real}
```
nickrobinson251: I am not sure `dim` is the right name for this, as it has a different meaning to the `dim` keyword as used throughout Julia Base and Stats packages

just calling it `d` seems less confusing to me. Maybe just `length`? although that has downside of shadowing the function. Perhaps we can come up with a more user-friendly name still?

src/grad.jl
```diff
@@ -1,40 +1,64 @@
...
+        gs = grad(fdm, x->f(x)[d], x)
+        for k in 1:length(x)
+            J[d, k] = gs[k]
+        end
     end
     return J
 end
-jacobian(fdm, f, x::Vector{<:Number}) = jacobian(fdm, f, x, length(f(x)))
+
+function jacobian(fdm, f, xs...; dim::Int=length(f(xs...)))
```
nickrobinson251: there is a lot of `...` splatting in the PR. I wonder (haven't checked) if it is worth using `Vararg{<:Any, N} where N`? (for performance https://github.com/JuliaLang/julia/pull/32817/files)

src/grad.jl
nickrobinson251: whatever we decide, docstring text needs updating

Roger-luo:
@willtebbutt  I can specialize it

```jl
function grad(fdm, f, d::Dict{K, V}) where {K, V}
    dd = Dict{K, V}()
    for (k, v) in d
        function f′(x)
            tmp = copy(d)
            tmp[k] = x
            return f(tmp)
        end
        dd[k] = grad(fdm, f′, v)
    end
    return dd
end
```

`Dict` is the only case that I didn't handle. I think the point is, allow specialize `grad` is already sufficient, no need to have an extra API.

src/grad.jl
Roger-luo: yeah, I agree. I'm good with any better name, I just used this randomly. I think `d` is too implicit, what do we usually call the `output`? maybe

`output_len`? or `len`?

src/grad.jl
Roger-luo: yeah, sure. revised.

src/grad.jl
nickrobinson251: sure :) `output_len`, `length` or even just `len` seem ok to me

src/grad.jl
Roger-luo: I'm not sure. I thought I didn't need to care much about performance in this package (since it was made for test). And I guess it would be fine, since places contains splatting are always tuple and only has a few lines.

Roger-luo: And I don't think I should use `Vararg{<:Any, N}` this will force all the args to be the same type, which is not correct. It is reasonable to allow functions like

```jl
f(rand(2, 2), 2)
```

to work with `grad` and `jacobian`

willtebbutt:
Hmm. Have you run tested for correctness on things like `Vector{<:AbstractVector}` and correctness tests for `Dict` with other element types?

Stepping back for a second: the point of `to_vec` is to separate out some problems. It's clear how to write `grad` and `jvp` et al to work on `Vector{<:Real}`s, so it makes sense to convert all other problems into problems on `Vector{<:Real}`s. For example, the benefit of using `to_vec` is that you don't need to implement `grad` for `Real`, `Tuple`, `Dict`, `Vector{<:AbstractVector}` etc, you just have to implement it for `Vector{<:Real}` and everything else via `to_vec` (while also handling multiple arguments, as you've done very nicely in this PR :) ). Then if someone encounters a new type that isn't properly handled, they just need to implement an appropriate `to_vec` method, and it will then become correctly handled. This is (arguably) a much simpler task than specialising `grad` to a new type. It's also great because `jvp` and `j′vp` also utilise it to handle non-`Vector{<:Real}` types, so if you add a new `to_vec` method, all three functions will benefit.

In essence what you're doing with your specialisations of `grad` for non-`Vector{<:Real}` types is to manually reimplement stuff that's already been done in `to_vec`.

My suggestion would be to let `to_vec` handle all of the specialisations, and just ensure that `grad` is correctly implemented for multiple arguments of type `Vector{<:Real}`.

Hopefully that's clear, let me know if not :)

src/grad.jl
nickrobinson251: all good points. I am convinced. thanks! :)

src/grad.jl
Roger-luo: I used `len`

Roger-luo:
> Have you run tested for correctness on things like Vector{<:AbstractVector} and correctness tests for Dict with other element types?

I didn't specialize `Vector{<:AbstractVector}`, but if the Dict's value are one of `<:Dict, Vector{<:Number}, <:Numer`, it can give the correct results. But you are right, I missed the type bound for grad, but it is not related to this issue.

> Stepping back for a second: the point of to_vec is to separate out some problems. It's clear how to write grad and jvp et al to work on Vector{<:Real}s, so it makes sense to convert all other problems into problems on Vector{<:Real}s. For example, the benefit of using to_vec is that you don't need to implement grad for Real, Tuple, Dict, Vector{<:AbstractVector} etc, you just have to implement it for Vector{<:Real} and everything else via to_vec

I think on the contrary, it can be more straight forward to specialize `grad` in some cases, and the performance slow down is pretty large and can be easily avoided because of the `to_vec` API. let's just take this small example in test: https://github.com/JuliaDiff/FiniteDifferences.jl/blob/master/test/grad.jl#L91 

```jl
julia> d = Dict(:a=>5, :b=>randn(10, 11), :c=>(5, 4, 3))
Dict{Symbol,Any} with 3 entries:
  :a => 5
  :b => [-1.02771 -0.481541 … 0.76612 -0.405744; -0.662716 0.676414 … 0.0109345 -0.678737…
  :c => (5, 4, 3)

julia> fdm1 = central_fdm(5, 1)
FiniteDifferenceMethod:
  order of method:       5
  order of derivative:   1
  grid:                  -2:2
  coefficients:          [0.0833333, -0.666667, 0.0, 0.666667, -0.0833333]


julia> function foo(d)
       sum(d[:a] .+ d[:b] .+ sum(d[:c]))
       end
foo (generic function with 1 method)

julia> @benchmark grad(fdm1, foo, $d)
BenchmarkTools.Trial:
  memory estimate:  2.82 MiB
  allocs estimate:  39303
  --------------
  minimum time:     2.701 ms (0.00% GC)
  median time:      2.911 ms (0.00% GC)
  mean time:        3.103 ms (5.36% GC)
  maximum time:     43.116 ms (91.45% GC)
  --------------
  samples:          1610
  evals/sample:     1

julia> function grad2(fdm, f, x)
           v, back = to_vec(x)
           return back(grad(fdm, x->f(back(v)), v))
       end
grad2 (generic function with 1 method)

julia> @benchmark grad2(fdm1, foo, $d)
BenchmarkTools.Trial:
  memory estimate:  6.70 MiB
  allocs estimate:  91330
  --------------
  minimum time:     6.821 ms (0.00% GC)
  median time:      7.667 ms (0.00% GC)
  mean time:        8.272 ms (7.08% GC)
  maximum time:     48.519 ms (82.49% GC)
  --------------
  samples:          604
  evals/sample:     1
``` 

`grad` is the specialized version in this PR. You should be able to run it with this branch. I haven't checked the `jvp` and `jacobian` but since they all make use of `grad`, they should be slowed down by `to_vec` a lot as well.


> It's also great because jvp and j′vp also utilise it to handle non-Vector{<:Real} types, so if you add a new to_vec method, all three functions will benefit.

I think I kinda get your point, these implementation will benefit from a `to_vec` implementation indeed, but they also suffer this performance issue due to this API(which is about 2x+). So I may propose to let `grad` fallback to `to_vec` version for now. So it won't break what's already there, but if anyone want to specialize it for better performance, that will also work.

Roger-luo:
BTW, should I add some sort of deprecation warning on the old `jacobian` API that has an optional arg `D` instead of the new kwarg?

willtebbutt:
> So I may propose to let grad fallback to to_vec version for now. So it won't break what's already there, but if anyone want to specialize it for better performance, that will also work.

This sounds good to me.

I definitely take your point regarding performance, but you're correct that we're less concerned about performance for this package than we are generality and keeping the maintenance required to as low a level as possible. That being said, faster is usually better, so having the option to specialise `grad` directly is nice.

> BTW, should I add some sort of deprecation warning on the old jacobian API that has an optional arg D instead of the new kwarg?

Good idea, this would be helpful.

test/grad.jl
```diff
@@ -44,6 +44,43 @@ Base.length(x::DummyType) = size(x.X, 1)
...
+        f3(x::Tuple) = sum(x[1]) + x[2]
+        dxs = grad(fdm, f3, (x, y))
+        @test dxs[1] ≈ grad(fdm, x->f3((x, y)), x)
+        @test dxs[2] ≈ grad(fdm, y->f3((x, y)), y)
+
+        f4(d::Dict) = sum(d[:x]) + d[:y]
+        d = Dict(:x=>x, :y=>y)
+        dxs = grad(fdm, f4, d)
+        @test dxs[:x] ≈ grad(fdm, x->f3((x, y)), x)
+        @test dxs[:y] ≈ grad(fdm, y->f3((x, y)), y)
```
nickrobinson251: another small style thing: can each of these blocks seperated by blank-lines just be there own (nested) testset? (I find testsets very helpful for maintainability)

Roger-luo: done.

nickrobinson251: thank you!

nickrobinson251:
LGTM

PR #34, opened 2019-08-04
Check that the derivative is nonnegative, and handle the case `q == 0`

### LINKED ISSUE #32: How do I choose method order to avoid NaN?
I am running into a situation where choosing specific orders for the Method is causing `NaN`s. Is this expected behaviour (e.g. resulting from some unfortunate math) or is this a bug?

In this example I am trying to take `n`th order directional derivatives. While silly, I've evaluated the `0`th order fdm, and on alternating method orders I'm getting `NaN`s:

```
@test sin(2.) ≈ central_fdm(1,0)(ϵ -> sin(2. +ϵ*1),0.) # Nan
@test sin(2.) ≈ central_fdm(2,0)(ϵ -> sin(2. +ϵ*1),0.) # true
@test sin(2.) ≈ central_fdm(3,0)(ϵ -> sin(2. +ϵ*1),0.) # Nan
@test sin(2.) ≈ central_fdm(4,0)(ϵ -> sin(2. +ϵ*1),0.) # true

@test cos(2.) ≈ central_fdm(2,1)(ϵ -> sin(2. +ϵ*1),0.) # true
@test cos(2.) ≈ central_fdm(3,1)(ϵ -> sin(2. +ϵ*1),0.) # true
@test cos(2.) ≈ central_fdm(4,1)(ϵ -> sin(2. +ϵ*1),0.) # true
@test cos(2.) ≈ central_fdm(5,1)(ϵ -> sin(2. +ϵ*1),0.) # true

@test -sin(2.) ≈ central_fdm(3,2)(ϵ -> sin(2. +ϵ*1),0.) # false but ok
@test -sin(2.) ≈ central_fdm(4,2)(ϵ -> sin(2. +ϵ*1),0.) # true
@test -sin(2.) ≈ central_fdm(5,2)(ϵ -> sin(2. +ϵ*1),0.) # true
@test -sin(2.) ≈ central_fdm(6,2)(ϵ -> sin(2. +ϵ*1),0.) # true
```

Note that I'm only observing this on the 0th order `fdm`.

wesselb:
Closes #32

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -145,7 +145,7 @@ end
 
 # Check the method and derivative orders for consistency
 function _check_p_q(p::Integer, q::Integer)
-    q > 0 || throw(ArgumentError("order of derivative must be strictly greater than zero"))
+    q >= 0 || throw(ArgumentError("order of derivative must be nonnegative"))
     q < p || throw(ArgumentError("order of the method must be strictly greater than that " *
                                  "of the derivative"))
     # Check whether the method can be computed. We require the factorial of the
@@ -225,6 +225,10 @@ function fdm(
     eps > 0 || throw(ArgumentError("eps must be positive, got $eps"))
     bound > 0 || throw(ArgumentError("bound must be positive, got $bound"))
     0 <= adapt < 20 - m.p || throw(ArgumentError("can't perform $adapt adaptation steps"))
+
+    # The below calculation can fail for `m.q == 0`, because then `C₂` may be zero. We
+    # therefore handle this edge case here.
+    m.q == 0 && return m, f(x)
 
     p = m.p
     q = m.q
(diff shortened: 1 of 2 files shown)


jessebett:
The default behaviour of `derivative(f,0)` should not be an assertion error, it should just return the function evaluation.

willtebbutt:
> The default behaviour of derivative(f,0) should not be an assertion error, it should just return the function evaluation.

I can see the appeal. @wesselb any thoughts?

wesselb:
Also sounds fair to me. @jessebett, how does the most commit look?

nickrobinson251:
FYI I will register FiniteDifferences.jl as a a new package with a new UUID and version 0.1 (because renaming a package is hard/confusing/undefined behaviour AFAIU) 

Should I wait for this PR before doing that?

PR #39, opened 2019-08-13
Faster jvp-s

willtebbutt:
We should never have been using the `jacobian` function to estimate jacobian-vector products. Instead, one should estimate the jvp directly by estimating the derivative in the direction of `ẋ`. It's definitely faster, and probably more numerically stable.

### DIFF
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -39,7 +39,7 @@ jacobian(fdm, f, x::Vector{<:Real}) = jacobian(fdm, f, x, length(f(x)))
 
 Convenience function to compute `jacobian(f, x) * ẋ`.
 """
-_jvp(fdm, f, x::Vector{<:Real}, ẋ::AV{<:Real}) = jacobian(fdm, f, x) * ẋ
+_jvp(fdm, f, x::Vector{<:Real}, ẋ::AV{<:Real}) = fdm(ε -> f(x .+ ε .* ẋ), zero(eltype(x)))
 
 """
     _j′vp(fdm, f, ȳ::AbstractVector{<:Real}, x::Vector{<:Real})


nickrobinson251:
Can you post some benchmarks? :)

nickrobinson251:
Can the `_j′vp` case also be improved?

willtebbutt:
> Can you post some benchmarks? :)

```julia
using BenchmarkTools, FiniteDifferences
using FiniteDifferences: jvp

A = randn(100, 100)
x = randn(100)
ẋ = randn(100)

f(x) = A * x

@benchmark jvp($(central_fdm(5, 1)), f, ($x, $ẋ))

# master:
julia> @benchmark jvp($(central_fdm(5, 1)), f, ($x, $ẋ))
BenchmarkTools.Trial: 
  memory estimate:  155.36 MiB
  allocs estimate:  2600310
  --------------
  minimum time:     274.782 ms (1.75% GC)
  median time:      277.619 ms (1.71% GC)
  mean time:        281.347 ms (2.82% GC)
  maximum time:     349.868 ms (17.44% GC)
  --------------
  samples:          18
  evals/sample:     1

# new branch:
julia> @benchmark jvp($(central_fdm(5, 1)), f, ($x, $ẋ))
BenchmarkTools.Trial: 
  memory estimate:  46.02 KiB
  allocs estimate:  263
  --------------
  minimum time:     29.758 μs (0.00% GC)
  median time:      31.145 μs (0.00% GC)
  mean time:        38.755 μs (15.75% GC)
  maximum time:     40.971 ms (99.76% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

> Can the _j′vp case also be improved?

Unfortunately not. It's the case that finite differencing essentially estimates forwards-mode AD, so we can exploit that to avoid computing the jacobian :)

Issue #22, opened 2019-06-19
Documentation


willtebbutt:
Our docs appear to be a bit out of date. In particular the `jvp` and `j′vp` stuff isn't well documented :(

willtebbutt:
Also, we're missing descriptions of `grad` and `jacobian` on the landing page. If someone could sort this out, that would be very helpful.

PR #42, opened 2019-08-19
Include `FiniteDifferenceMethod` and gradient functions in docs

### LINKED ISSUE #41: Include `FiniteDifferenceMethod` types in docs
One part of #22 

AFAIK these are part of the pubilc API

- My preferred solution is to just export them (currently we doument all and only exports)

- Alternatively, in case e.g. `Forward` and `Backward` are too generic as names to export, we can just add them in their own `@doc` block

nickrobinson251:
- now exported and in docs
- closes #41

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@
 *.jl.mem
 docs/build/
 docs/site/
-Manifest.toml
+/Manifest.toml
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,6 @@ jobs:
       julia: 1.0
       os: linux
       script:
-        - julia --project=docs/ -e 'using Pkg; Pkg.instantiate(); Pkg.develop(PackageSpec(path=pwd()))'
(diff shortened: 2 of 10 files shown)


nickrobinson251:
How about
- `grad`
- `jacobian`
- `jvp`
- `j′vp`
- `to_vec`
should these be exported and/or included in the docs? 
(#22 suggests at least include them in the docs?)

nickrobinson251:
Not sure if `docs/Manifest.toml` is the preferred method or not? 

Was it better with `using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()` and no Manifest?

nickrobinson251:
I copied LibPQ.jl on the Manifest.toml thing - so it can't be completely wrong 😆 

BTW, adding exports (as this does) is usually considered breaking - it could cause a name clash in some existing code - so any release including these changes will need to be v0.8

PR #43, opened 2019-08-31
Relax type constraints to allow complex jvp

simeonschaub:
This is needed for tests in JuliaDiff/ChainRules.jl#95.

### DIFF
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -1,7 +1,7 @@
 export grad, jacobian, jvp, j′vp, to_vec
 
 """
-    grad(fdm, f, x::AbstractVector)
+    grad(fdm, f, x::Vector{<:Number})
 
 Approximate the gradient of `f` at `x` using `fdm`. Assumes that `f(x)` is scalar.
 """
@@ -21,8 +21,8 @@ function grad(fdm, f, x::Vector{T}) where T<:Number
 end
 
 """
-    jacobian(fdm, f, x::AbstractVector{<:Real}, D::Int)
-    jacobian(fdm, f, x::AbstractVector{<:Real})
+    jacobian(fdm, f, x::AbstractVector{<:Number}, D::Int)
+    jacobian(fdm, f, x::AbstractVector{<:Number})
 
 Approximate the Jacobian of `f` at `x` using `fdm`. `f(x)` must be a length `D` vector. If
 `D` is not provided, then `f(x)` is computed once to determine the output size.
@@ -37,16 +37,16 @@ end
 jacobian(fdm, f, x::Vector{<:Number}) = jacobian(fdm, f, x, length(f(x)))
 
 """
-    _jvp(fdm, f, x::Vector{<:Real}, ẋ::AbstractVector{<:Real})
+    _jvp(fdm, f, x::Vector{<:Number}, ẋ::AbstractVector{<:Real})
 
 Convenience function to compute `jacobian(f, x) * ẋ`.
 """
 _jvp(fdm, f, x::Vector{<:Number}, ẋ::AV{<:Number}) = fdm(ε -> f(x .+ ε .* ẋ), zero(eltype(x)))
 
 """
-    _j′vp(fdm, f, ȳ::AbstractVector{<:Real}, x::Vector{<:Real})
+    _j′vp(fdm, f, ȳ::AbstractVector{<:Number}, x::Vector{<:Number})
 
-Convenience function to compute `jacobian(f, x)' * ȳ`.
+Convenience function to compute `transpose(jacobian(f, x)) * ȳ`.
 """
 _j′vp(fdm, f, ȳ::AV{<:Number}, x::Vector{<:Number}) = transpose(jacobian(fdm, f, x, length(ȳ))) * ȳ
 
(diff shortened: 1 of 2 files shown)


src/grad.jl
```diff
@@ -27,28 +27,28 @@ end
...
-jacobian(fdm, f, x::Vector{<:Real}) = jacobian(fdm, f, x, length(f(x)))
+jacobian(fdm, f, x::Vector{<:Number}) = jacobian(fdm, f, x, length(f(x)))
 
 """
     _jvp(fdm, f, x::Vector{<:Real}, ẋ::AbstractVector{<:Real})
 
 Convenience function to compute `jacobian(f, x) * ẋ`.
 """
-_jvp(fdm, f, x::Vector{<:Real}, ẋ::AV{<:Real}) = fdm(ε -> f(x .+ ε .* ẋ), zero(eltype(x)))
+_jvp(fdm, f, x::Vector{<:Number}, ẋ::AV{<:Number}) = fdm(ε -> f(x .+ ε .* ẋ), zero(eltype(x)))
```
nickrobinson251: All the docstrings will need changing too :)

src/grad.jl
```diff
@@ -27,28 +27,28 @@ end
...
 
 """
     _j′vp(fdm, f, ȳ::AbstractVector{<:Real}, x::Vector{<:Real})
 
 Convenience function to compute `jacobian(f, x)' * ȳ`.
 """
-_j′vp(fdm, f, ȳ::AV{<:Real}, x::Vector{<:Real}) = jacobian(fdm, f, x, length(ȳ))' * ȳ
+_j′vp(fdm, f, ȳ::AV{<:Number}, x::Vector{<:Number}) = transpose(jacobian(fdm, f, x, length(ȳ))) * ȳ
```
nickrobinson251: Can the `Real` -> `Number` change be done for the entire file? Currently `Real` will remain in a few places

src/grad.jl
```diff
@@ -5,7 +5,7 @@ export grad, jacobian, jvp, j′vp, to_vec
 
 Approximate the gradient of `f` at `x` using `fdm`. Assumes that `f(x)` is scalar.
 """
-function grad(fdm, f, x::Vector{T}) where T<:Real
+function grad(fdm, f, x::Vector{T}) where T<:Number
```
nickrobinson251: I suppose really we'd need to add some tests on `Complex` cases if we're going to support them?

src/grad.jl
simeonschaub: For now, I only changed the constraints, where I actually needed too. I think it would be safer for now to leave the other constraints and only loosen them when needed in ChainRules.jl

src/grad.jl
simeonschaub: Oops, yes!

src/grad.jl
nickrobinson251: Hmm, yeah, I do see that point of view

It's important to make sure FiniteDifferences itself has consistent behaviour e.g. it is surprising for `to_vec` to work on a Complex number but not an array of Complex numbers

But 🤷‍♂ let's see what Will or Wessel think

src/grad.jl
simeonschaub: Yes probably. Though it would be nice, if we could just do something analogous to JuliaDiff/ChainRulesCore.jl#28 and not having to write tests twice :wink:

src/grad.jl
simeonschaub: `to_vec` seems harmless enough to correctly handle `Vector{<:Number}`, so I've added that definition.

simeonschaub: I also suppose, it would make more sense to rename `j′vp` to `jᵀvp`, but that's would definitely be breaking and out of the scope of this PR

willtebbutt:
Thanks for this @simeonschaub. Would you mind adding some tests involving complex numbers to `to_vec` tests? It might be overkill, but better safe than sorry.

simeonschaub:
@willtebbutt Definitely not overkill, since I discovered that `to_vec` for `Adjoint` wasn't defined correctly for complex arrays.

test/grad.jl
```diff
@@ -55,38 +55,51 @@ Base.length(x::DummyType) = size(x.X, 1)
...
-            test_to_vec((randn(4), randn(4, 3, 2), 1))
-            test_to_vec((5, randn(4, 3, 2), UpperTriangular(randn(4, 4)), 2.5))
-            test_to_vec(((6, 5), 3, randn(3, 2, 0, 1)))
-            test_to_vec((DummyType(randn(2, 7)), DummyType(randn(3, 9))))
-            test_to_vec((DummyType(randn(3, 2)), randn(11, 8)))
-        end
-        @testset "Dictionary" begin
-            test_to_vec(Dict(:a=>5, :b=>randn(10, 11), :c=>(5, 4, 3)))
+    for T in (Float64, ComplexF64)
+        @testset "to_vec(::$T)" begin
```
oxinabox: Ratherr than a `for`-loop + a `@testset` we can use the testset form that combines them:
```
@testset "to_vec(::$T)"  for T in (Float64, ComplexF64)
```

simeonschaub:
Would it be possible to tag this right away?

willtebbutt:
On it

PR #44, opened 2019-09-01
CI Changes / Improvements

willtebbutt:
This PR ensure that CI runs on all relevant versions of Julia, with allowed failure for 1.3 and nightly (obviously the former will have to change once 1.3 is released).

It also moves our windows CI from appveyor to travis. This just makes maintenance slightly more straightforward.


.travis.yml
```diff
@@ -3,14 +3,19 @@ language: julia
 os:
   - linux
   - osx
+  - windows
 julia:
   - 0.7
   - 1.0
+  - 1.1
```
nickrobinson251: Let's not bother with 0.7 and 1.1, since they're not maintained?

nickrobinson251:
Also I think we need to change the Julia compat in Project.toml to `julia = "1"` to allow any 1.x version

.travis.yml
oxinabox: Keep 1.1,
drop 0.7

willtebbutt: > Keep 1.1, for now.

What's the reason for this? Seems as good a plan as any to drop support for versions at the same rate as the language.

PR #51, opened 2019-10-04
change return type of grad/jacobian to match args

Roger-luo:
I didn't notice this yesterday until I start working on the Zygote PR today. So the basically current return value type of single variable grad/jacobian is not very consistent with multivariate grad/jacobian. It should return `(x, )`, this is causing extra handling when I implement `gradcheck` using the jacobians in Zygote.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.8.0"
+version = "0.9.0"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -94,7 +94,7 @@ is defined.
 function jvp(fdm, f, (x, ẋ)::Tuple{Any, Any})
     x_vec, vec_to_x = to_vec(x)
     _, vec_to_y = to_vec(f(x))
-    return (vec_to_y(_jvp(fdm, x_vec->to_vec(f(vec_to_x(x_vec)))[1], x_vec, to_vec(ẋ)[1])), )
+    return vec_to_y(_jvp(fdm, x_vec->to_vec(f(vec_to_x(x_vec)))[1], x_vec, to_vec(ẋ)[1]))
 end
 function jvp(fdm, f, xẋs::Tuple{Any, Any}...)
     x, ẋ = collect(zip(xẋs...))
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -94,11 +94,11 @@ is defined.
 function jvp(fdm, f, (x, ẋ)::Tuple{Any, Any})
     x_vec, vec_to_x = to_vec(x)
     _, vec_to_y = to_vec(f(x))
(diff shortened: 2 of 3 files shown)


nickrobinson251:
I will defer to WIll on this.

If this is what is mostly widely useful (and not surprising) behaviour in practice, then i'm okay with it. But we should be cobfident it is before making the change.

Also this is a breaking change, so at the very least it'd be good to bump the version number accordingly in this PR.

(As a complete aside I'm really looking forward to being able to unwrap singe-element collections with `only(f(x))` instead of `f(x)[1]` in Julia 1.4+)

Roger-luo:
@nickrobinson251 I can change it to `x[]` instead of `x[1]` if you prefer that. But practically I think I need this for Zygote or I have to implement an extra dispatch in Zygote that deal with single input. But notice this is also not consistent with `Zygote`'s `gradient`.

nickrobinson251:
> practically I think I need this for Zygote or I have to implement an extra dispatch in Zygote that deal with single input. But notice this is also not consistent with Zygote's gradient.

Seem like compelling reasons to me! And the code looks good. But I'd like @willtebbutt before going ahead.

(Also, yeah, `[1]` is good here. And as you say we won't be able to use `only`, unless we add it to Comat.jl or whatever, it's just that cases like this remind me that I wish we had it / glad we will one day.)

willtebbutt:
I'm pro- this change, but we need to enforce it for all functions, including `jvp` and `j′vp`. So if you could make those changes as well, that would be great.

willtebbutt:
@Roger-luo any chance you're able to finish this off? It should only require minor modifications to `jvp` and `j′vp` to make this work as intended, and the changes would be analogous to the ones that you've already made.

Would be great to get this merged so that we can get [this PR](https://github.com/FluxML/Zygote.jl/pull/235) merged and resolve [this](https://github.com/FluxML/Zygote.jl/issues/283) and other related issues.

Roger-luo:
Yeah, sorry I was having mid term exams... I'll have time on this in the weekend I think. @willtebbutt

willtebbutt:
Fantastic, thanks @Roger-luo . Hope the mid terms went okay!

Roger-luo:
@willtebbutt just applied this to `jvp` and `j'vp` I think, `jvp` always returns a single number so just `(x, )` would be sufficient?

willtebbutt:
If @oxinabox is happy, I'm happy to merge.

oxinabox:
I have not reviewed but I trust @willtebbutt who knows this code base much better than I do

willtebbutt:
Actually, we definitely want `j′vp` to return a `Tuple`, but I don't think that we want `jvp` to since there's not need to handle multiple outputs in the same way. We should probably revert to a single output, rather than a `Tuple`. Apologies for the confusion!

Roger-luo:
@willtebbutt is this good?

willtebbutt:
Yup, good by me. Will merge once tests have passed.

Roger-luo:
BTW I'll need a tag to make CI in Zygote pass when this is merged later I think

willtebbutt:
Understood. Could you bump to 0.9.0 in this PR?

PR #54, opened 2019-11-22
faster and more compatible jvp for Zygote

Roger-luo:
I think this change of `_jvp` won't effect the API, will it? just some modification to use multi-args jacobian directly

### DIFF
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -48,16 +48,18 @@ function grad(fdm, f, xs...)
     end
 end
 
+_length_of_output(f, xs...) = length(to_vec(f(xs...))[1])
+
 """
     jacobian(fdm, f, xs::Union{Real, AbstractArray{<:Real}}; len::Int=length(f(x)))
 
 Approximate the Jacobian of `f` at `x` using `fdm`. `f(x)` must be a length `len` vector. If
 `len` is not provided, then `f(x)` is computed once to determine the output size.
 """
-function jacobian(fdm, f, x::Union{T, AbstractArray{T}}; len::Int=length(f(x))) where {T <: Number}
+function jacobian(fdm, f, x::Union{T, AbstractArray{T}}; len::Int=_length_of_output(f, x)) where {T <: Number}
     J = Matrix{float(T)}(undef, len, length(x))
     for d in 1:len
-        gs = grad(fdm, x->f(x)[d], x)[1]
+        gs = grad(fdm, x->to_vec(f(x))[1][d], x)[1]
         for k in 1:length(x)
             J[d, k] = gs[k]
         end
@@ -66,17 +68,17 @@ function jacobian(fdm, f, x::Union{T, AbstractArray{T}}; len::Int=length(f(x))) 
 end
 
 # if not number or array
-function jacobian(fdm, f, x; len::Int=length(f(x)))
+function jacobian(fdm, f, x; len::Int=_length_of_output(f, x))
     vec_x, vec_to_x = to_vec(x)
     return jacobian(fdm, x->f(vec_to_x(x)), vec_x; len=len)
 end
 
 # tuple should return tuple
-function jacobian(fdm, f, x::Tuple; len::Int=length(f(x)))
+function jacobian(fdm, f, x::Tuple; len::Int=_length_of_output(f, x))
     return (jacobian(fdm, (xs...)->f(tuple(xs...)), x...; len=len), )
 end
 
-function jacobian(fdm, f, xs...; len::Int=length(f(xs...)))
+function jacobian(fdm, f, xs...; len::Int=_length_of_output(f, xs...))
     return ntuple(length(xs)) do k
         jacobian(fdm, x->f(replace_arg(x, xs, k)...), xs[k]; len=len)[1]
     end
@@ -127,7 +129,7 @@ end
 
 Compute an adjoint with any types of arguments for which [`to_vec`](@ref) is defined.
 """
-function j′vp(fdm, f, ȳ, xs...; len::Int=length(f(xs...)))
+function j′vp(fdm, f, ȳ, xs...; len::Int=_length_of_output(f, xs...))
     Js = jacobian(fdm, f, xs...; len=len)
     return Tuple(_j′vp(J, ȳ, x) for (J, x) in zip(Js, xs))
 end
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -71,6 +71,11 @@ function jacobian(fdm, f, x; len::Int=length(f(x)))
     return jacobian(fdm, x->f(vec_to_x(vec_x)), vec_x; len=len)
 end
 
+# tuple should return tuple
+function jacobian(fdm, f, x::Tuple; len::Int=length(f(x)))
+    (jacobian(fdm, (xs...)->f(tuple(xs...)), x...; len=len), )
+end
+
 function jacobian(fdm, f, xs...; len::Int=length(f(xs...)))
     return ntuple(length(xs)) do k
         jacobian(fdm, x->f(replace_arg(x, xs, k)...), xs[k]; len=len)[1]
@@ -93,6 +98,12 @@ function _j′vp(jac, ȳ, x)
     Δ, vec_to_y = to_vec(ȳ)
     _, vec_to_x = to_vec(x)
     return vec_to_x(transpose(jac) * Δ)
+end
+
+function _j′vp(jac::Tuple, ȳ, x::Tuple)
+    ntuple(length(x)) do k
+        _j′vp(jac[k], ȳ, x[k])
+    end
 end
 
 """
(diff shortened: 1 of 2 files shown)


src/grad.jl
```diff
@@ -106,12 +110,10 @@ end
 
 Compute an adjoint with any types of arguments for which [`to_vec`](@ref) is defined.
 """
-function j′vp(fdm, f, ȳ, x)
-    x_vec, vec_to_x = to_vec(x)
-    ȳ_vec, _ = to_vec(ȳ)
-    return (vec_to_x(_j′vp(fdm, x_vec->to_vec(f(vec_to_x(x_vec)))[1], ȳ_vec, x_vec)), )
+function j′vp(fdm, f, ȳ, xs...; len::Int=length(f(xs...)))
+    Js = jacobian(fdm, f, xs...; len=len)
+    Tuple(_j′vp(J, ȳ, x) for (J, x) in zip(Js, xs))
```
oxinabox:
```suggestion
    return Tuple(_j′vp(J, ȳ, x) for (J, x) in zip(Js, xs))
```

src/grad.jl
```diff
@@ -79,11 +79,15 @@ Convenience function to compute `jacobian(f, x) * ẋ`.
...
-    _j′vp(fdm, f, ȳ::AbstractVector{<:Number}, x::Vector{<:Number})
+    _j′vp(jacobian, ȳ, x)
 
 Convenience function to compute `transpose(jacobian(f, x)) * ȳ`.
 """
-_j′vp(fdm, f, ȳ::AV{<:Number}, x::Vector{<:Number}) = transpose(jacobian(fdm, f, x; len=length(ȳ))[1]) * ȳ
+function _j′vp(jac, ȳ, x)
+    Δ, vec_to_y = to_vec(ȳ)
+    _, vec_to_x = to_vec(x)
+    vec_to_x(transpose(jac) * Δ)
```
oxinabox:
```suggestion
    return vec_to_x(transpose(jac) * Δ)
```

willtebbutt:
These changes look okay to me. Would be good to understand why the tests are currently failing though.

Roger-luo:
Oh there are some tests on the old _j'vp will it be fine to just remove it?

oxinabox:
I can't review this package.
Ideally @wesselb would.

willtebbutt:
Have added reviewing this to my calendar for Monday, as I'm not going to have time to look at this before then.

src/grad.jl
```diff
@@ -65,6 +65,17 @@ function jacobian(fdm, f, x::Union{T, AbstractArray{T}}; len::Int=length(f(x)))
...
 
+# if not number or array
+function jacobian(fdm, f, x; len::Int=length(f(x)))
+    vec_x, vec_to_x = to_vec(x)
+    return jacobian(fdm, x->f(vec_to_x(vec_x)), vec_x; len=len)
+end
+
+# tuple should return tuple
+function jacobian(fdm, f, x::Tuple; len::Int=length(f(x)))
+    (jacobian(fdm, (xs...)->f(tuple(xs...)), x...; len=len), )
```
oxinabox:
```suggestion
    return (jacobian(fdm, (xs...)->f(tuple(xs...)), x...; len=len), )
```

willtebbutt:
@Roger-luo Firstly, apologies for taking ages to review this -- I've been completely swamped.

I like the idea of the PR, but there's a thing that doesn't appear to work that I would like to ensure keeps working.
```julia
f(x) = (5x, 4x)
x = randn(5)
ȳ = (randn(5), randn(5))
j′vp(central_fdm(5, 1), f, ȳ, x)
```
doesn't currently function correctly in this PR as is makes  a call to `_j′vp` without first ensuring that the inputs are correctly `to_vec`-ed, which would in turn ensure that `jacobian` operates properly. The idea is to ensure that `_j′vp` only ever sees vector-like things. So the solution is almost certainly to do some appropriate `to_vec`ing in your new `j′vp` implementation, in a similar manner to the current implementation.

I also appreciate that this particular example isn't currently tested, so I would grateful if you could add a test based on this to your PR :)

Roger-luo:
no problem. I didn't consider tuple as j'vp at first. I'll add this later after my finals lol. Before that when you got time would you mind helping me on the Zygote gradcheck PR (https://github.com/FluxML/Zygote.jl/pull/235)? I still can't figure out why eigvals test is not passing  :-(

src/grad.jl
```diff
@@ -65,6 +65,17 @@ function jacobian(fdm, f, x::Union{T, AbstractArray{T}}; len::Int=length(f(x)))
     return (J, )
 end
 
+# if not number or array
+function jacobian(fdm, f, x; len::Int=length(f(x)))
+    vec_x, vec_to_x = to_vec(x)
+    return jacobian(fdm, x->f(vec_to_x(vec_x)), vec_x; len=len)
```
wesselb: Should this be `x->f(vec_to_x(x))` here?

Roger-luo: oh yeah, nice catch!

nickrobinson251:
Bump on this :) Would also be good to add tests to keep code coverage up -- thanks again for your work on this @Roger-luo!

Roger-luo:
Oops sorry, I forgot this PR. I'll look at next week.

oxinabox:
bump

Roger-luo:
@willtebbutt your example doesn't work here because I'm using `jacobian` to calculate the Jacobian matrix and `jacobian` doesn't work for multi-return functions, the original implementation of jvp converted multi-return function's return values to a vector using `to_vec`, and this is something never considered in Zygote or other reverse mode AD packages I believe, I've posted an issue to discuss this, I think I prefer a more structured solution but we'll see. 

I'll keep the current master behaviour in this PR for now (convert it to a vector). We could change it if on Zygote side we are going to use a more structured jacobian API.

willtebbutt:
> @willtebbutt your example doesn't work here because I'm using jacobian to calculate the Jacobian matrix and jacobian doesn't work for multi-return functions, the original implementation of jvp converted multi-return function's return values to a vector using to_vec, and this is something never considered in Zygote or other reverse mode AD packages I believe

I don't believe this is true. `Zygote.pullback` can certainly handle functions that return `Tuple`s, so we definitely need to be able to handle them.

Again, apologies for taking ages to work on this. Upon further reflection, I'm a bit worried about this PR -- it really messes with the core design principle behind `FiniteDifferences` which is

1. "flatten" everything into a `Vector{<:Real}` (non-Vector{<:Real} things, multiple-arguments)
2. compute stuff (jacobian, jvp, jpvp, grad etc)
3. unflatten where possible

This last step is possible for all operations _except_ Jacobian, because we don't currently have a way of representing `Jacobian`s between arbitrary objects anywhere in the Julia ecosystem. So we have a workaround in the multi-args case that's something of a special case.

> I think I prefer a more structured solution but we'll see.

I would ultimately like to see this too, but it's not the solution that we have available to us at the minute. Not even `ChainRules` has started thinking about how to handle this, so it's definitely a medium- to long-term thing. The above approach is designed precisely to ensure composability -- all you have to be able to do is ensure that `to_vec` is defined on the arguments to your function and you should be grand.

What I'm lacking at the minute is an understanding of precisely how these changes achieve the desired outcome. To that end, it would be helpful if you could provide

1. a demonstration of how this improves performance. This would be helpful because it's not immediately clear that this going to happen often / that it will be substantial.
2. a concrete example of the problem this solves for Zygote's usage. Again, it's not completely clear to me how what we've currently got doesn't work with Zygote -- in any case I would have anticipated that it would need to use `jpvp` rather than `jacobian`.

I definitely agree that our `jacobian` function needs some attention -- it was never really designed to be a user-facing function (jvp / jpvp are the main things we expect users to use, and possibly grad). For example, the type constraints on the method that does most of the work are currently far too broad imho, and its structure should be broken up to better reflect the interface for jvp and jpvp.

Again, apologies for bringing these concerns up so late in the day -- I'm not sure why they didn't occur to me last time I looked at this PR.

Roger-luo:
> I don't believe this is true. Zygote.pullback can certainly handle functions that return Tuples, so we definitely need to be able to handle them.

Yeah, I mean if we want to handle this in a consistent way comparing to Zygote's `pullback` we will need structured Jacobian where for multi-return functions the Jacobian should be in a Tuple rather than a large matrix.

> Again, apologies for taking ages to work on this. Upon further reflection, I'm a bit worried about this PR -- it really messes with the core design principle behind FiniteDifferences which is

This is not your case, your case is a multi-return function, which is different from multivariant function, their structured Jacobian type is different.

> What I'm lacking at the minute is an understanding of precisely how these changes achieve the desired outcome. To that end, it would be helpful if you could provide

I think the main issue was that the API was not consistent to Zygote, which blocks the gradcheck PR on Zygote side. I see your new refactor, it looks good, I'll close this first and try your new version and see if that works. Thanks for the review.

PR #59, opened 2020-02-01
Do gradient via mutating and unmutating cell

oxinabox:
We can avoid copying every element every step if we just mutate the element we want to mutate on the original matrix, then mutate it back again after so we still have the original for the next step.

My micro benchmarks show this helps a bit.

### DIFF
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -8,18 +8,24 @@ Approximate the gradient of `f` at `xs...` using `fdm`. Assumes that `f(xs...)` 
 """
 function grad end
 
-function grad(fdm, f, x::AbstractArray{T}) where T <: Number
+function _grad(fdm, f, x::AbstractArray{T}) where T <: Number
+    # x must be mutable, we will mutate it and then mutate it back.
     dx = similar(x)
-    tmp = similar(x)
     for k in eachindex(x)
         dx[k] = fdm(zero(T)) do ϵ
-            tmp .= x
-            tmp[k] += ϵ
-            return f(tmp)
+            xk = x[k]
+            x[k] = xk +  ϵ
+            ret = f(x)
+            x[k] = xk  # Can't do `x[k] -= ϵ` as floating-point math is not associative
+            return ret
         end
     end
     return (dx, )
 end
+
+grad(fdm, f, x::Array{<:Number}) = _grad(fdm, f, x)
+# Fallback for when we don't know `x` will be mutable:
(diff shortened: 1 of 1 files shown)


oxinabox:
I think similar could be done for the `Dict` case but I don't really understand how that works.
https://github.com/JuliaDiff/FiniteDifferences.jl/pull/59/files#diff-94e14cf43e653c4b17f49cc18895018cR33-R44

longemen3000:
with the dict:
```
function grad(fdm, f, d::Dict{K, V}) where {K, V}
    ∇d = Dict{K, V}()
    for (k, v) in d
        temp = d[k]
        function f′(x)
            d[k] = x
            return f(d)
        end
        ∇d[k] = grad(fdm, f′, v)[1]
        d[k] = temp
    end
    return (dd, )
end

```

oxinabox:
### Benchmarks

```
using BenchmarkTools
using FiniteDifferences

f(x) = sum(sin, x);
@btime grad($_fdm, $f, $xs);
```


# Before:
  222.774 ms (154790 allocations: 4.28 MiB)

# With this PR;
  212.711 ms (150807 allocations: 4.20 MiB)

So only a very modest improvement.
But still a step towards not allocating at all (beyond what the primal does).

src/grad.jl
```diff
@@ -8,31 +8,38 @@ Approximate the gradient of `f` at `xs...` using `fdm`. Assumes that `f(xs...)`
...
-            tmp[k] = x
-            return f(tmp)
+            d[k] = x
+            return f(d)
         end
-        dd[k] = grad(fdm, f′, v)[1]
+        ∇d[k] = grad(fdm, f′, v)[1]
+        d[k] = dk
     end
     return (dd, )
```
oxinabox:
```suggestion
    return (∇d, )
```

src/grad.jl
```diff
@@ -8,33 +8,40 @@ Approximate the gradient of `f` at `xs...` using `fdm`. Assumes that `f(xs...)`
 """
 function grad end
 
-function grad(fdm, f, x::AbstractArray{T}) where T <: Number
+function _grad(fdm, f, x::AbstractArray{T}) where T <: Number
```
willtebbutt: Doesn't technically matter, but maybe change / remove the type constraint on `x`, since we know it's going to be an `Array` by virtue of the fact it's called from `grad`

oxinabox: Could be a a few things actually.
But yes, the constraints are not needed

oxinabox: I am going to keep it because the `T` is useful

PR #60, opened 2020-02-02
Decrease allocations

oxinabox:
Turns out type-stability was causing a bunch of allocations.
Type inference was failing for some reason.

Also tracking history surpisingly, causes allocations.
We don't need to track history when doing grad anyway.

This is built ontop of #59  and that should be merged first.

```
using BenchmarkTools
using FiniteDifferences

const _fdm5 = central_fdm(5,1);

f(x) = sum(sin, x);
@btime grad($_fdm, $f, $xs);
@btime grad($_fdm5, $f, $xs);
```

This PR decreases allocation count by >60%
and decreases allocation amoun t by >30%.

But that does not translate directly into speed.
I guess it won't pay off much til the garbage collector gets avoided...

### Current (oxmut1)
```
julia> @btime grad($_fdm, $f, $xs);
  202.552 ms (154789 allocations: 4.26 MiB)

julia> @btime grad($_fdm5, $f, $xs);
  413.311 ms (162753 allocations: 5.45 MiB)
```
-----------------

### New with Stability improvements
```
julia> @btime grad($_fdm, $f, $xs);
  200.205 ms (49777 allocations: 2.66 MiB)

julia> @btime grad($_fdm5, $f, $xs);
  398.832 ms (59732 allocations: 3.87 MiB)
```

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -181,6 +181,7 @@ The recognized keywords are:
 * `bound`: Bound on the value of the function and its derivatives at `x`.
 * `condition`: The condition number. See [`DEFAULT_CONDITION`](@ref).
 * `eps`: The assumed roundoff error. Defaults to `eps()` plus [`TINY`](@ref).
+* `track_history`: wether to update the history of the method `m` with e.g. accuracy stats.
 
 !!! warning
     Bounds can't be adaptively computed over nonstandard grids; passing a value for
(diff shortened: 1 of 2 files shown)


src/methods.jl
```diff
@@ -181,6 +181,7 @@ The recognized keywords are:
 * `bound`: Bound on the value of the function and its derivatives at `x`.
 * `condition`: The condition number. See [`DEFAULT_CONDITION`](@ref).
 * `eps`: The assumed roundoff error. Defaults to `eps()` plus [`TINY`](@ref).
+* `track_history`: wether to update the history of the method `m` with e.g. accuracy stats.
```
willtebbutt:
```suggestion
* `track_history`: whether to update the history of the method `m` with e.g. accuracy stats.
```

willtebbutt:
@oxinabox what's the status of this PR?

oxinabox:
Current status is the test failures need debugging and fixing

PR #61, opened 2020-02-02
Precompute grids and coeffs

oxinabox:
grids and coeffs are always the same for standard grids for given `p` and `q`.

Precomputing them  helps decrease the cost assoicated with `adapt` in the gradients.
This also fixes up some of the type instability with the central gradient, during adapt.

```
using BenchmarkTools
using FiniteDifferences

const _fdm5 = central_fdm(5,1);

f(x) = sum(sin, x);
@btime grad($_fdm, $f, $xs);
@btime grad($_fdm5, $f, $xs);
```

### Current
```
julia> @btime grad($_fdm, $f, $xs);
  202.552 ms (154789 allocations: 4.26 MiB)

julia> @btime grad($_fdm5, $f, $xs);
  413.311 ms (162753 allocations: 5.45 MiB)
```
-----------------

### New with Precomputed coeffs
```
julia> @btime grad($_fdm, $f, $xs);
  195.610 ms (53759 allocations: 1.96 MiB)

julia> @btime grad($_fdm5, $f, $xs);
  386.412 ms (53759 allocations: 1.87 MiB)
```

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -138,14 +138,15 @@ for D in (:Forward, :Backward, :Central)
         const $precomputes = ntuple($num_precomputed) do p
             p == 1 && return nothing  # no grid define for p==1
             grid = $gridf(p)
-            coefs = _coefs(grid, p, 1)
-            (grid, coefs)
+            all_coefs = ntuple(q->_coefs(grid, p, q), p - 1)
+            (grid, all_coefs)
         end
 
         function $D(p::Integer, q::Integer; adapt=1, kwargs...)
             _check_p_q(p, q)
-            if p <= $num_precomputed && q == 1
-                grid, coefs = $precomputes[p]
+            if p <= $num_precomputed
+                grid, all_coefs = $precomputes[p]
+                coefs = all_coefs[q]
             else
                 grid = $gridf(p)
                 coefs = _coefs(grid, p, q)
@@ -197,6 +198,7 @@ The recognized keywords are:
 * `bound`: Bound on the value of the function and its derivatives at `x`.
 * `condition`: The condition number. See [`DEFAULT_CONDITION`](@ref).
 * `eps`: The assumed roundoff error. Defaults to `eps()` plus [`TINY`](@ref).
+* `track_history`: wether to update the history of the method `m` with e.g. accuracy stats.
 
 !!! warning
     Bounds can't be adaptively computed over nonstandard grids; passing a value for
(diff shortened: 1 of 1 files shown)


oxinabox:
From stacking this ontop of #60:
```
julia> @btime grad($_fdm, $f, $xs);
  193.883 ms (51768 allocations: 1.93 MiB)

julia> @btime grad($_fdm5, $f, $xs);
  400.359 ms (51768 allocations: 1.84 MiB)
```

With #60 alone:
```
julia> @btime grad($_fdm, $f, $xs);
  200.205 ms (49777 allocations: 2.66 MiB)

julia> @btime grad($_fdm5, $f, $xs);
  398.832 ms (59732 allocations: 3.87 MiB)
```

interesting that it seems to make the `central_fdm(2)` case slightly worse.
Not sure whats up with that.

oxinabox:
From stacking this ontop of #60:
```
julia> @btime grad($_fdm, $f, $xs);
  193.883 ms (51768 allocations: 1.93 MiB)

julia> @btime grad($_fdm5, $f, $xs);
  400.359 ms (51768 allocations: 1.84 MiB)
```

With #60 alone:
```
julia> @btime grad($_fdm, $f, $xs);
  200.205 ms (49777 allocations: 2.66 MiB)

julia> @btime grad($_fdm5, $f, $xs);
  398.832 ms (59732 allocations: 3.87 MiB)
```

interesting that it seems to make the `central_fdm(2)` case slightly worse.
Not sure whats up with that.

KristofferC:
Instead of precomputing you could memoize. Avoids having to determine upfront what exact combinations should be precomputed. Taking the cost once per session is probably ok.

src/methods.jl
```diff
@@ -97,20 +122,35 @@ end
...
 
+    num_precomputed = 8
     @eval begin
         # Compatibility layer over the "old" API
         function $fdmf(p::Integer, q::Integer; adapt=1, kwargs...)
             _dep_kwarg(kwargs)
             return $D(p, q; adapt=adapt, kwargs...)
         end
 
+        # precompute a bunch of grids and coefs for the q=1 case
```
oxinabox:
```suggestion
        # precompute a bunch of grids and coefs
```

src/methods.jl
```diff
@@ -97,20 +122,35 @@ end
...
+    num_precomputed = 8
     @eval begin
         # Compatibility layer over the "old" API
         function $fdmf(p::Integer, q::Integer; adapt=1, kwargs...)
             _dep_kwarg(kwargs)
             return $D(p, q; adapt=adapt, kwargs...)
         end
 
+        # precompute a bunch of grids and coefs
+        const $precomputes = ntuple($num_precomputed) do p
```
oxinabox: Should this be defined in a constant?
Or should it be declared outside the eval, and then inlined in as a local literal
and thus should constant fold

oxinabox:
> Instead of precomputing you could memoize. Avoids having to determine upfront what exact combinations should be precomputed. Taking the cost once per session is probably ok.

Yeah, good idea.

I think I might have looked into it, and struggled to workout how to do it without getting allocations.
The precomputes use tuples which are very compiler friendly.

But I can't recall if I did, so I should check again

PR #65, opened 2020-03-28
Compute coeffs using Rationals

### LINKED ISSUE #64: Less accurate results on Julia 1.3.1 vs Julia 1.1.1
If I run
```
FiniteDifferences.fdm(FiniteDifferences.central_fdm(9, 5), y->exp(y), 1.0, adapt=4)
```
on Julia 1.1.1 I get:
```
2.718281560120131
```
whereas on Julia 1.3.1 I get
```
2.7182639618941993
```
which is much less accurate (error of 1.8e-5 vs 2.7e-7).

This broke some tests of ours. Does anyone know why the accuracy is reduced?

Both on v0.9.2.

oxinabox:
Fixed #64 

Turns out this can make a difference.

This not only gets result on julia 1.4 up to the standard of julia 1.2,
but actually does much more than that, significantly improving out accuracy

### With this PR. in julia 1.2 or 1.4 (identical):
```julia
julia> using FiniteDifferences

julia> k3 = FiniteDifferences.fdm(FiniteDifferences.central_fdm(9, 5), y->exp(y), 1.0, adapt=4)
2.7182817496168528

julia> abs(k3-exp(1))  # Should be 0
7.884219233034173e-8

julia> f(x) = @. 4x^3 + 3x^2 + 2x + 1
f (generic function with 1 method)

julia> FiniteDifferences.fdm(FiniteDifferences.central_fdm(9, 3), f, 1.0, adapt=4) - 24
5.538680625249981e-12
```

## Without this PR. 1.4
```julia
julia> using FiniteDifferences

julia> k3 = FiniteDifferences.fdm(FiniteDifferences.central_fdm(9, 5), y->exp(y), 1.0, adapt=4)
2.7182663156747684

julia> abs(k3-exp(1))  # Should be 0
1.5512784276694447e-5

julia> f(x) = @. 4x^3 + 3x^2 + 2x + 1
f (generic function with 1 method)

julia> FiniteDifferences.fdm(FiniteDifferences.central_fdm(9, 3), f, 1.0, adapt=4) - 24
6.48253006829691e-10
```

### Without this PR on 1.2
```julia
julia> using FiniteDifferences

julia> k3 = FiniteDifferences.fdm(FiniteDifferences.central_fdm(9, 5), y->exp(y), 1.0, adapt=4)
2.7182815825355453

julia> abs(k3-exp(1))  # Should be 0
2.459234997864712e-7

julia> f(x) = @. 4x^3 + 3x^2 + 2x + 1
f (generic function with 1 method)

julia> FiniteDifferences.fdm(FiniteDifferences.central_fdm(9, 3), f, 1.0, adapt=4) - 24
1.071143174158351e-11
```

---

I feel like this is going to increase our cost a fair bit.
So it would be really good to get #61 done, to avoid recomputing coeffs.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.9.3"
+version = "0.9.4"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -157,10 +157,12 @@ end
 
 # Compute coefficients for the method
 function _coefs(grid::AbstractVector{<:Real}, p::Integer, q::Integer)
-    C = [g^i for i in 0:(p - 1), g in grid]
-    x = zeros(Int, p)
+    # For high precision on the \ we use Rational, and to prevent overfloats we use Int128
+    # At the end we go to Float64 for fast floating point math (rather than rational math)
+    C = [Rational{Int128}(g^i) for i in 0:(p - 1), g in grid]
+    x = zeros(Rational{Int128}, p)
     x[q + 1] = factorial(q)
-    return C \ x
+    return Float64.(C \ x)
 end
 
 # Estimate the bound on the function value and its derivatives at a point
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -157,7 +157,7 @@ end
 
 # Compute coefficients for the method
 function _coefs(grid::AbstractVector{<:Real}, p::Integer, q::Integer)
-    # For high precision on the \ we use Rational, and to prevent overfloats we use Int128
+    # For high precision on the \ we use Rational, and to prevent overflows we use Int128
     # At the end we go to Float64 for fast floating point math (rather than rational math)
     C = [Rational{Int128}(g^i) for i in 0:(p - 1), g in grid]
     x = zeros(Rational{Int128}, p)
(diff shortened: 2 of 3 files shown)


willtebbutt:
This looks like a fantastic improvement in accuracy -- I'm definitely in favour of doing this. Would definitely be good to prevent reverting by tying this down with a couple of unit tests. The above snippets would probably suffice.

oxinabox:
I think the improvements are only really big when using higher order and a high degree of `adapt`

willtebbutt:
> I think the improvements are only really big when using higher order and a high degree of adapt

Oh okay. Would be good to see some numbers on this then, both benchmarking and accuracy.

oxinabox:
tests have been added

src/methods.jl
```diff
@@ -157,10 +157,12 @@ end
 
 # Compute coefficients for the method
 function _coefs(grid::AbstractVector{<:Real}, p::Integer, q::Integer)
-    C = [g^i for i in 0:(p - 1), g in grid]
-    x = zeros(Int, p)
+    # For high precision on the \ we use Rational, and to prevent overfloats we use Int128
```
willtebbutt, line 160:
```suggestion
    # For high precision on the \ we use Rational, and to prevent overflows we use Int128
```

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.9.2"
+version = "0.9.3"
```
oxinabox, line 3:
```suggestion
version = "0.9.4"
```

PR #66, opened 2020-03-29
Refactor jacobian and grad

willtebbutt:
This is a fairly substantial PR, motivated in no small part by @Roger-luo 's work in #54 .

### Overview

In particular it

1. completely refactors `jacobian` for performance reasons. Our previous implementation was really quite bad (my fault) -- it required ~~O(input_dims^2)~~ O(input_dims * output_dims) function evaluations rather than O(input_dims).
2. refactors `grad` to utilise `jpvp`. `grad` is now a single-line function that utilises the new `jacobian` implementation. This is a substantial drop in code complexity with no loss of functionality, as much of the code that we had before to handle dictionaries etc is completely gone. The result is that `grad` now depends on `to_vec` like everything else in this package.
3. the new implementation of `jacobian` has the pleasing advantage of doing away with the need to pass a `len` parameter around. This is a breaking change, hence the minor version bump.
4. `to_vec` has been moved out of `grad.jl` to emphasise the separation of concerns.

### Benchmarking

```julia
using BenchmarkTools, FiniteDifferences

A = randn(15, 10)
B = randn(10, 12)

@benchmark jacobian($(central_fdm(5, 1)), *, $A, $B)

# master
BenchmarkTools.Trial:
  memory estimate:  1.02 GiB
  allocs estimate:  4520890
  --------------
  minimum time:     637.381 ms (2.42% GC)
  median time:      654.539 ms (2.73% GC)
  mean time:        660.101 ms (2.80% GC)
  maximum time:     691.597 ms (3.33% GC)
  --------------
  samples:          8
  evals/sample:     1

# New code
BenchmarkTools.Trial:
  memory estimate:  15.97 MiB
  allocs estimate:  45663
  --------------
  minimum time:     5.207 ms (0.00% GC)
  median time:      5.380 ms (0.00% GC)
  mean time:        5.627 ms (4.77% GC)
  maximum time:     8.298 ms (13.32% GC)
  --------------
  samples:          888
  evals/sample:     1
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.9.2"
+version = "0.10.0"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.10.0"
+version = "0.9.3"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -6,7 +6,8 @@ Approximate the Jacobian of `f` at `x` using `fdm`. Results will be returned as 
 version of `x`, and `y_vec` the flattened version of `f(x...)`. Flattening performed by
 [`to_vec`](@ref).
 """
-function jacobian(fdm, f, x::Vector{<:Number})
+function jacobian(fdm, f, x::Vector{<:Number}; len=nothing)
+    len !== nothing && Base.depwarn("len parameter to jacobian is deprecated", :jacobian)
     ẏs = map(eachindex(x)) do n
         return fdm(zero(eltype(x))) do ε
             xn = x[n]
@@ -19,14 +20,14 @@ function jacobian(fdm, f, x::Vector{<:Number})
     return (hcat(ẏs...), )
 end
 
-function jacobian(fdm, f, x)
+function jacobian(fdm, f, x; len=nothing)
     x_vec, from_vec = to_vec(x)
-    return jacobian(fdm, f ∘ from_vec, x_vec)
+    return jacobian(fdm, f ∘ from_vec, x_vec; len=len)
 end
 
-function jacobian(fdm, f, xs...)
+function jacobian(fdm, f, xs...; len=nothing)
     return ntuple(length(xs)) do k
-        jacobian(fdm, x->f(replace_arg(x, xs, k)...), xs[k])[1]
+        jacobian(fdm, x->f(replace_arg(x, xs, k)...), xs[k]; len=len)[1]
     end
 end
 
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -6,7 +6,7 @@ export grad, jacobian, jvp, j′vp
 Approximate the Jacobian of `f` at `x` using `fdm`. Results will be returned as a
 `Matrix{<:Number}` of `size(length(y_vec), length(x_vec))` where `x_vec` is the flattened
 version of `x`, and `y_vec` the flattened version of `f(x...)`. Flattening performed by
-`to_vec`.
+[`to_vec`](@ref).
 """
 function jacobian(fdm, f, x::Vector{<:Number})
     ẏs = map(eachindex(x)) do n
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -81,4 +81,4 @@ j′vp(fdm, f, ȳ, xs...) = j′vp(fdm, xs->f(xs...), ȳ, xs)[1]
 
 Compute the gradient of `f` for any `xs` for which [`to_vec`](@ref) is defined.
 """
-grad(fdm, f, xs...) = j′vp(fdm, f, 1, xs...)
+grad(fdm, f, xs...) = j′vp(fdm, f, 1, xs...)  # `j′vp` with seed of 1
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -59,60 +59,13 @@ end
 """
     grad(fdm, f, xs...)
 
-Approximate the gradient of `f` at `xs...` using `fdm`. Assumes that `f(xs...)` is scalar.
+Approximate the gradient of `f` at `xs...` using `fdm`. Requires that `f(xs...)` is scalar.
 """
-function grad end
-
-function _grad(fdm, f, x::AbstractArray{T}) where T <: Number
-    # x must be mutable, we will mutate it and then mutate it back.
-    dx = similar(x)
-    for k in eachindex(x)
-        dx[k] = fdm(zero(T)) do ϵ
-            xk = x[k]
-            x[k] = xk + ϵ
-            ret = f(x)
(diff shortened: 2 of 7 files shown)


src/to_vec.jl
```diff
@@ -0,0 +1,100 @@
+export to_vec
```
oxinabox: this export should be in the main file.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.9.2"
+version = "0.10.0"
```
oxinabox: What makes this change breaking?

src/grad.jl
```diff
@@ -1,96 +1,47 @@
...
-            x[k] = xk +  ϵ
-            ret = f(x)
-            x[k] = xk  # Can't do `x[k] -= ϵ` as floating-point math is not associative
+function jacobian(fdm, f, x::Vector{<:Number})
+    ẏs = map(eachindex(x)) do n
+        return fdm(zero(eltype(x))) do ε
+            xn = x[n]
+            x[n] = xn + ε
+            ret = (first ∘ to_vec ∘ f)(x)
+            x[n] = xn
```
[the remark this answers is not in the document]
nickrobinson251: i liked the old comment ```# Can't do `x[k] -= ϵ` as floating-point math is not associative```

src/grad.jl
```diff
@@ -1,96 +1,47 @@
...
-_jvp(fdm, f, x::Vector{<:Number}, ẋ::AV{<:Number}) = fdm(ε -> f(x .+ ε .* ẋ), zero(eltype(x)))
-
-"""
-    _j′vp(fdm, f, ȳ::AbstractVector{<:Number}, x::Vector{<:Number})
-
-Convenience function to compute `transpose(jacobian(f, x)) * ȳ`.
-"""
-_j′vp(fdm, f, ȳ::AV{<:Number}, x::Vector{<:Number}) = transpose(jacobian(fdm, f, x; len=length(ȳ))[1]) * ȳ
+function _jvp(fdm, f, x::Vector{<:Number}, ẋ::Vector{<:Number})
+    return fdm(ε -> f(x .+ ε .* ẋ), zero(eltype(x)))
```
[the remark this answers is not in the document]
willtebbutt: It's not clear to be that we couldn't have a complex `x` and real `ẋ`, so maybe better not to do this?

Project.toml
willtebbutt: The change of the `jacobian` API. If someone is currently using the `len` kwarg, that will now break.

src/grad.jl
```diff
@@ -113,108 +64,21 @@ end
 
 Compute an adjoint with any types of arguments for which [`to_vec`](@ref) is defined.
 """
+function _j′vp(fdm, f, ȳ::Vector{<:Number}, x::Vector{<:Number})
```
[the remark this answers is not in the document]
willtebbutt: I agree that we should leave this for a follow up PR. We can include it in the minor release bump from this PR though.

src/grad.jl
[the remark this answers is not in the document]
oxinabox: ok good

Project.toml
oxinabox: can we just deprecate it, rather than break it?

willtebbutt: We could, but all it'll do is completely ignore the `len` parameter. Is that something we want? (I'm not overly familiar with the conventions here)

oxinabox: Yeah, that's not a breaking change.
Because there is only 1 valid value of `len` and passing anything else is an error anyway.
So  a deprecation that ignores `len` is nonbreaking

may have do do a manual deprecatation inside the `jacobian` function using `if` and `Base.depwarn`

willtebbutt: Done. If the emitted warning is appropriate, please resolve the conversation.

src/to_vec.jl
```diff
@@ -0,0 +1,100 @@
...
+function to_vec(X::Diagonal)
+    function Diagonal_from_vec(x_vec)
+        return Diagonal(reshape(x_vec, size(X)...))
+    end
+    return vec(Matrix(X)), Diagonal_from_vec
+end
+
+function to_vec(X::Transpose)
+    function Transpose_from_vec(x_vec)
+        return Transpose(permutedims(reshape(x_vec, size(X))))
```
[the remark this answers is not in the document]
willtebbutt: Yeah, I think we're good. Pretty sure this is the right thing.

willtebbutt:
@oxinabox thanks for the above comments. They're now resolved other than the deprecation, which I would appreciate feedback on.

src/grad.jl
```diff
@@ -1,96 +1,46 @@
...
-    # x must be mutable, we will mutate it and then mutate it back.
-    dx = similar(x)
-    for k in eachindex(x)
-        dx[k] = fdm(zero(T)) do ϵ
-            xk = x[k]
-            x[k] = xk +  ϵ
-            ret = f(x)
-            x[k] = xk  # Can't do `x[k] -= ϵ` as floating-point math is not associative
+function jacobian(fdm, f, x::Vector{<:Number}; len=nothing)
+    len !== nothing && Base.depwarn("len parameter to jacobian is deprecated", :jacobian)
```
oxinabox: I would tweak the wording a bit. (which unfortunately puts it over the line limit)
```suggestion
    len !== nothing && Base.depwarn(
        "`len` keyword argument to `jacobian` is nonlonger required " *
        "and will not be permitted in the future.",
         :jacobian
    )
```

Other than that looks perfect.

oxinabox:
> They're now resolved other than the deprecation, which I would appreciate feedback on

Have given feedback.

Ideally would wait for @Roger-luo  to give feedback as to if this worksout, before we merge this

src/grad.jl
nickrobinson251: typo in suggestion: "non" -> "no "

src/grad.jl
```diff
@@ -1,96 +1,50 @@
...
-    dx = similar(x)
-    for k in eachindex(x)
-        dx[k] = fdm(zero(T)) do ϵ
-            xk = x[k]
-            x[k] = xk +  ϵ
-            ret = f(x)
-            x[k] = xk  # Can't do `x[k] -= ϵ` as floating-point math is not associative
+function jacobian(fdm, f, x::Vector{<:Number}; len=nothing)
+    len !== nothing && Base.depwarn(
+        "`len` keyword argument to `jacobian` is nolonger required " *
```
nickrobinson251:
```suggestion
        "`len` keyword argument to `jacobian` is no longer required " *
```

willtebbutt: Damn it, could've sworn I have fixed that! Thanks for catching it.

Roger-luo:
Yeah, this seems to work well, thanks!

willtebbutt:
@oxinabox @wesselb @nickrobinson251 any further comments on this?

src/grad.jl
```diff
@@ -1,96 +1,50 @@
...
+        "`len` keyword argument to `jacobian` is no longer required " *
+        "and will not be permitted in the future.",
+         :jacobian
+    )
+    ẏs = map(eachindex(x)) do n
+        return fdm(zero(eltype(x))) do ε
+            xn = x[n]
+            x[n] = xn + ε
+            ret = first(to_vec(f(x)))
+            x[n] = xn  # Can't do `x[k] -= ϵ` as floating-point math is not associative
```
wesselb, line 20:
```suggestion
            x[n] = xn  # Can't do `x[n] -= ϵ` as floating-point math is not associative
```

Issue #67, opened 2020-04-08
Breaking changes in v0.9.3


sethaxen:
v0.9.3 introduced some breaking changes

Setup:
```julia
julia> using LinearAlgebra, FiniteDifferences, StaticArrays
julia> x = Float64[1, 2, 3]
julia> fdm = central_fdm(5, 1)
julia> x2 = SVector{3}(x);
```

### Jacobian of the identity function no longer produces the identity matrix.
```julia
julia> jacobian(fdm, identity, x)[1] # expected identity matrix
```

on v0.1.2:
```julia
3×3 Array{Float64,2}:
 1.0          6.50515e-15  6.50515e-15
 1.30103e-14  1.0          1.30103e-14
 0.0          0.0          1.0
```

on v0.1.3:
```julia
3×3 Array{Float64,2}:
 4.16334e-16  4.16334e-16  4.16334e-16
 8.32667e-16  8.32667e-16  8.32667e-16
 0.0          0.0          0.0
```

### StaticArrays no longer supported

```julia
julia> jacobian(fdm, identity, x2)[1]
```

on v0.1.2:
```julia
 3×3 Array{Float64,2}:
  1.0          6.50515e-15  6.50515e-15
  1.30103e-14  1.0          1.30103e-14
  0.0          0.0          1.0
```

on v0.1.3:
```julia
ERROR: MethodError: no method matching to_vec(::SArray{Tuple{3},Float64,1,3})
Closest candidates are:
  to_vec(::Number) at /Users/saxen/.julia/packages/FiniteDifferences/kqb1h/src/to_vec.jl:8
  to_vec(::Array{#s121,1} where #s121<:Number) at /Users/saxen/.julia/packages/FiniteDifferences/kqb1h/src/to_vec.jl:15
  to_vec(::Array{T,1} where T) at /Users/saxen/.julia/packages/FiniteDifferences/kqb1h/src/to_vec.jl:17
  ...
Stacktrace:
 [1] jacobian(::FiniteDifferences.Central{UnitRange{Int64},Array{Float64,1}}, ::Function, ::SArray{Tuple{3},Float64,1,3}; len::Nothing) at /Users/saxen/.julia/packages/FiniteDifferences/kqb1h/src/grad.jl:28
[...]

PR #68, opened 2020-04-08
Fixes for #67

willtebbutt:
This is a partial resolution for #67 @sethaxen - there's the still the `StaticArrays` issue, that requires a bit more work.

I'm not sure if this is going to the be correct solution forever, but it's a decent patch for now.

### DIFF
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -16,7 +16,7 @@ function jacobian(fdm, f, x::Vector{<:Number}; len=nothing)
         return fdm(zero(eltype(x))) do ε
             xn = x[n]
             x[n] = xn + ε
-            ret = first(to_vec(f(x)))
+            ret = copy(first(to_vec(f(x))))
             x[n] = xn  # Can't do `x[n] -= ϵ` as floating-point math is not associative
             return ret
         end
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -4,9 +4,9 @@
 Transform `x` into a `Vector`, and return the vector, and a closure which inverts the
 transformation.
 """
-function to_vec(x::T) where {T<:Number}
+function to_vec(x::Number)
     function Number_from_vec(x_vec)
-        return T(first(x_vec))
+        return oftype(x, first(x_vec))
     end
     return [x], Number_from_vec
 end
@@ -14,24 +14,24 @@ end
 # (Abstract)Vectors
 to_vec(x::Vector{<:Number}) = (x, identity)
 
-function to_vec(x::T) where {T<:AbstractVector}
+function to_vec(x::AbstractVector)
     x_vecs_and_backs = map(to_vec, x)
     x_vecs, backs = first.(x_vecs_and_backs), last.(x_vecs_and_backs)
     function Vector_from_vec(x_vec)
         sz = cumsum(map(length, x_vecs))
         x_Vec = [backs[n](x_vec[sz[n] - length(x_vecs[n]) + 1:sz[n]]) for n in eachindex(x)]
-        return T(x_Vec)
+        return oftype(x, x_Vec)
     end
     return vcat(x_vecs...), Vector_from_vec
 end
 
 # (Abstract)Arrays
-function to_vec(x::T) where {T<:AbstractArray}
+function to_vec(x::AbstractArray)
 
     x_vec, from_vec = to_vec(vec(x))
 
     function Array_from_vec(x_vec)
-        return T(reshape(from_vec(x_vec), size(x)))
+        return oftype(x, reshape(from_vec(x_vec), size(x)))
     end
 
     return x_vec, Array_from_vec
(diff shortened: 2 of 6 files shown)


test/grad.jl
```diff
@@ -52,6 +52,9 @@ using FiniteDifferences: grad, jacobian, _jvp, jvp, j′vp, _j′vp, to_vec
         @test Ac == A
         check_jac_and_jvp_and_j′vp(fdm, x->sin.(A * x), ȳ, x, ẋ, cos.(A * x) .* A)
         @test Ac == A
+
+        # Prevent regression against 67.
```
oxinabox, line 56:
```suggestion
        # Prevent regression against https://github.com/JuliaDiff/FiniteDifferences.jl/issues/67
```

src/grad.jl
```diff
@@ -16,7 +16,7 @@ function jacobian(fdm, f, x::Vector{<:Number}; len=nothing)
         return fdm(zero(eltype(x))) do ε
             xn = x[n]
             x[n] = xn + ε
-            ret = first(to_vec(f(x)))
+            ret = copy(first(to_vec(f(x))))
```
oxinabox, line 19:
```suggestion
            ret = copy(first(to_vec(f(x))))  # copy required incase `f(x)` returns something that aliases `x`
```

willtebbutt:
This now fixes all of #67 , and I'm pretty confident that the changes aren't breaking. There are a bunch of extra tests to make sure that the `from_vec` bit of `to_vec` returns precisely the right type, and to ensure that `SArray`s play nicely with `to_vec`.

These extra tests threw up a couple of bugs that (somehow) hadn't been causing problems, but were definitely bugs. These bugs have been fixed.

src/to_vec.jl
```diff
@@ -4,42 +4,41 @@
 Transform `x` into a `Vector`, and return the vector, and a closure which inverts the
 transformation.
 """
-function to_vec(x::Number)
+function to_vec(x::T) where {T<:Number}
     function Number_from_vec(x_vec)
-        return first(x_vec)
+        return T(first(x_vec))
```
oxinabox, line 9: You may like `oftype(x, first(x_vec))` fo this, and below

src/to_vec.jl
```diff
@@ -4,42 +4,41 @@
...
-function to_vec(x::Vector)
+
+function to_vec(x::T) where {T<:AbstractVector}
     x_vecs_and_backs = map(to_vec, x)
     x_vecs, backs = first.(x_vecs_and_backs), last.(x_vecs_and_backs)
     function Vector_from_vec(x_vec)
         sz = cumsum(map(length, x_vecs))
-        return [backs[n](x_vec[sz[n] - length(x_vecs[n]) + 1:sz[n]]) for n in eachindex(x)]
+        x_Vec = [backs[n](x_vec[sz[n] - length(x_vecs[n]) + 1:sz[n]]) for n in eachindex(x)]
+        return T(x_Vec)
```
oxinabox, line 23: Here i think it is more important to use `oftype` because it calls out to `convert` which is more appropriate than just assuming a constructor would exist taking this arg.
```suggestion
        return oftype(x, x_Vec)
```

Because `oftype` using `convert`

src/to_vec.jl
```diff
@@ -4,42 +4,41 @@
...
+# (Abstract)Arrays
+function to_vec(x::T) where {T<:AbstractArray}
+
+    x_vec, from_vec = to_vec(vec(x))
 
-function to_vec(x::Array)
-    x_vec, back = to_vec(reshape(x, :))
     function Array_from_vec(x_vec)
-        return reshape(back(x_vec), size(x))
+        return T(reshape(from_vec(x_vec), size(x)))
```
oxinabox, line 34: here also `oftype`

```suggestion
        return oftype(x, reshape(from_vec(x_vec), size(x)))
```

sethaxen:
> @sethaxen I would appreciate it if you could test this branch against Manifolds to check that it's fixed your problems? If not I'll figure out what other tests we should be adding.

I tested this branch against our master, and the test no longer fails. Thanks for the quick fix!

willtebbutt:
Thanks for checking!

@oxinabox I've implemented your `oftype` suggestions, but didn't merge your suggestions directly because there were related changes that needed to happen for code-cleanliness as a result.

src/to_vec.jl
```diff
@@ -6,40 +6,39 @@ transformation.
...
     function Number_from_vec(x_vec)
-        return first(x_vec)
+        return oftype(x, first(x_vec))
     end
     return [x], Number_from_vec
 end
 
-# Vectors
+# (Abstract)Vectors
 to_vec(x::Vector{<:Number}) = (x, identity)
```
nickrobinson251, line 15: why is this one still `Vector` not `AbstractVector`? Can we have a comment for it?

willtebbutt: Good point. Let me know if the new commenting clarifies things.

nickrobinson251: perfect! thanks :)

PR #71, opened 2020-04-11
Revert some of 0.9.4

willtebbutt:
As it turns out I was over-zealous with the type checking. Currently if you provide an `Int` input when computing differences, `to_vec` will make it a float if needs be. While it's debatable whether or not this is desirable behaviour, it's clearly breaking if it changes.

The changes to `adjoint` / `transpose` were just wrong, and they've now been reverted and a test added.

@oxinabox I would appreciate it greatly if you could verify whether these changes make all of the `ChainRules` tests pass on your machine. For some reason, but machine is throwing very strange errors in places where CI doesn't have them. I'm not sure whether it's an issue with my Julia install, my computer is dying (it could very well be, it's old and has been behaving a little strangely recently), something else that's weird, or an actual issue with FiniteDifferences.

### DIFF
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -61,25 +61,17 @@ function to_vec(X::Diagonal)
 end
 
 function to_vec(X::Transpose)
-
-    x_vec, x_from_vec = to_vec(X.parent)
-
     function Transpose_from_vec(x_vec)
-        return Transpose(x_from_vec(x_vec))
+        return Transpose(permutedims(reshape(x_vec, size(X))))
     end
-
-    return x_vec, Transpose_from_vec
+    return vec(Matrix(X)), Transpose_from_vec
 end
 
 function to_vec(X::Adjoint)
-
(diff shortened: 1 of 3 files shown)


src/to_vec.jl
```diff
@@ -61,25 +61,17 @@ function to_vec(X::Diagonal)
 end
 
 function to_vec(X::Transpose)
-
-    x_vec, x_from_vec = to_vec(X.parent)
-
     function Transpose_from_vec(x_vec)
-        return Transpose(x_from_vec(x_vec))
+        return Transpose(permutedims(reshape(x_vec, size(X))))
```
oxinabox, line 65: Why did this have to be reverted?

willtebbutt: It turned out that if you just look at the `parent` field, `first(to_vec(X)) == first(to_vec(X'))`, which as it turns out, is bad. The moral of the story was not to look at private fields, and to take the semantics of the matrix seriously.

PR #74, opened 2020-04-16
Fix travis to build docs

oxinabox:


oxinabox:
config validator passed so i think this will work

PR #85, opened 2020-05-19
Make to_vec output real for complex structured matrices

sethaxen:
Fixes #84

PR #86, opened 2020-05-19
Reconstruct Symmetric using uplo, and generalize for Hermitian also

sethaxen:
The `to_vec` function for `Hermitian` uses the same approach as used in #85.

### DIFF
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -55,7 +55,7 @@ end
 
 function to_vec(x::Symmetric)
     function Symmetric_from_vec(x_vec)
-        return Symmetric(reshape(x_vec, size(x)))
+        return Symmetric(reshape(x_vec, size(x)), Symbol(x.uplo))
     end
     return vec(Matrix(x)), Symmetric_from_vec
 end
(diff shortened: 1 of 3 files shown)


willtebbutt:
I guess we merge #85 then this. Please bump the patch version to 0.10.2

src/to_vec.jl
```diff
@@ -55,11 +55,19 @@ end
 
 function to_vec(x::Symmetric)
     function Symmetric_from_vec(x_vec)
-        return Symmetric(reshape(x_vec, size(x)))
+        return Symmetric(reshape(x_vec, size(x)), Symbol(x.uplo))
```
nickrobinson251, line 58: Why do we need the Symbol conversion here (same below)?

sethaxen: Even though `uplo` is internally a char, the `Symmetric(...)` constructor is only defined for `uplo::Symbol`. The alternative to this approach is to use the constructor with types: `typeof(x)(..., x.uplo)`.

nickrobinson251: Oh, wild. That seems odd, but maybe there's a reason for it. Thanks for explaining 😊

sethaxen: I switched over to the type constructor.

sethaxen: Yeah, there's some strangeness around `Symmetric` and `Hermitian`. 🤷‍♂️ No problem!

sethaxen:
Can this also be merged?

sethaxen:
Thanks!

PR #91, opened 2020-07-05
Implement rand_tangent and difference

willtebbutt:
This PR lays some foundations for (hopefully) moving `FiniteDifferences` over to `ChainRules` types, and contains exactly two things:

1. Moves `rand_tangent` over from `ChainRulesTestUtils`. It turns out that it's useful here, and since `ChainRulesTestUtils` depends on this package, there shouldn't be any harm implementing them here.
2. Implements `difference`, which is more or less what you might expect. If the docstring doesn't make it clear, it would be good to know.

edit: Also makes docs build on the latest Julia version.

edit-2: Also changes versions on which this is tested to `1.0` and `1`, and disallows 1.3 failure on nightly. I'm guessing this just hadn't been updated for some reason.


test/difference.jl
```diff
@@ -0,0 +1,57 @@
+function test_difference(ε::Real, x, dx)
+    y = x + ε * dx
+    dx_diff = difference(ε, y, x)
+    @test typeof(dx) == typeof(dx_diff)
```
willtebbutt, line 4: This testing is a bit weak. Ideally we would test that the definition in the docstring holds. `==` is out of the question, and most types don't have `isapprox` defined, so it's hard to know what to do. Suggestions appreciated.

test/difference.jl
```diff
@@ -0,0 +1,57 @@
...
+end
+
+@testset "difference" begin
+
+    @testset "$(typeof(x))" for (ε, x) in [
```
willtebbutt, line 9: These tests cover everything covered in the `to_vec` tests other than `Dict`, which doesn't currently play nicely / I'm not entirely sure how best to construct a tangent, nor how to add it to the primal.

willtebbutt:
CI failure is weird, pretty sure it's a one-off.

src/difference.jl
```diff
@@ -0,0 +1,45 @@
...
+type piracy while `-` and `/` cannot.
+"""
+difference(::Real, ::T, ::T) where {T<:Symbol} = DoesNotExist()
+difference(::Real, ::T, ::T) where {T<:AbstractChar} = DoesNotExist()
+difference(::Real, ::T, ::T) where {T<:AbstractString} = DoesNotExist()
+difference(::Real, ::T, ::T) where {T<:Integer} = DoesNotExist()
+
+difference(ε::Real, y::T, x::T) where {T<:Number} = (y - x) / ε
+
+difference(ε::Real, y::T, x::T) where {T<:StridedArray} = difference.(ε, y, x)
```
nickrobinson251, line 21: why does this need to be restricted to `StridedArray` rather than `AbstractArray`?

test/rand_tangent.jl
```diff
@@ -0,0 +1,85 @@
...
+   b::Int
+   c::Any
+end
+
+@testset "generate_tangent" begin
+    rng = MersenneTwister(123456)
+
+    foreach([
+
+        # Things without sensible tangents.
```
nickrobinson251, line 13: I think it would be more. idiomatic if we used a `for` loop (rather than `foreach(...) 

I also think the these tests would be clearer if we turned these comments into testsets e.g. something along the lines of
```julia
@testset "rand_tangent" begin
    rng = MersenneTwister(123456)
   
    @testset  "No tangent: $(typeof(x))" for x in ("hi", 'a', ...)
        @test rand_tangent(rng, x) isa DoesNotExist()
        @test rand_tangent(x) isa DoesNotExist()
        @test x + rand_tangent(rng, x) isa typeof(x)
    end

    @testset "Numbers: $(typeof(x))" for x in (5.0, 5.0 + 0.4im)
        @test rand_tangent(rng, x) isa typeof(x)
        @test rand_tangent(x) isa typeof(x)
        @test x + rand_tangent(rng, x) isa typeof(x)
    end  
    
    @testset "etc..."
end
```
Obviously something like this is more verbose, but I think it would be clearer

test/difference.jl
nickrobinson251: what types here cause a problem for `isapprox`?

presumably we should be okay to define `isapprox(::Zero, ::Zero) = true` if that's all we need?

test/rand_tangent.jl
```diff
@@ -0,0 +1,85 @@
+# Test struct for `rand_tangent`.
+struct Foo
```
nickrobinson251, line 2: this is also used in ` test/difference.jl` - shall we move it to `runtests.jl`?

src/difference.jl
willtebbutt: So that things like `Diagonal`s etc are handled by the generic struct interface, so as to prevent the creation of dense tangents for e.g. `Diagonal`s.

test/rand_tangent.jl
willtebbutt: Hmm I'd rather avoid a lot of copy + pasted code. Would something like:
```julia

    @testset "Primal: $(typeof(x)), Tangent: $T_tangent" for (x, T_tangent) in [

        # Things without sensible tangents.
        ("hi", DoesNotExist),
        ('a', DoesNotExist),
        (:a, DoesNotExist),
        (true, DoesNotExist),
        (4, DoesNotExist),
        ...
    ]
        @test rand_tangent(rng, x) isa T_tangent
        @test rand_tangent(x) isa T_tangent
        @test x + rand_tangent(rng, x) isa typeof(x)
    end

```
be acceptable?

willtebbutt:
> what types here cause a problem for isapprox?

For some reason I can't reply to this comment directly. It's certainly an issue for all composites, but purely in the sense that we've not defined it.

test/difference.jl
willtebbutt: Well it's certainly an issue for all composites, but purely in the sense that we've not defined it. We could define it. Maybe that's a good idea?

willtebbutt: I've opened https://github.com/JuliaDiff/ChainRulesCore.jl/issues/184 about this.

src/difference.jl
nickrobinson251: https://github.com/JuliaDiff/ChainRules.jl/issues/232

test/rand_tangent.jl
nickrobinson251: yes, i think that's an improvement as it'll give us a more informative message if something fails. thanks

willtebbutt:
Looks like the Manifest probably just needed re-generating -- it uses a relative path to get the correct version of `FiniteDifferences`. Does anyone know if this is the right way to do things (still?)?

oxinabox:
It is afaik.

test/difference.jl
nickrobinson251: Thanks. Can you also add a TODO to the code?
```suggestion
    # TODO: `@test isapprox(dx, dx_diff)` once `isapprox` is defined appropriately
    # see https://github.com/JuliaDiff/ChainRulesCore.jl/issues/184
    @test typeof(dx) == typeof(dx_diff)
```

willtebbutt:
@nickrobinson251 this PR no longer exports `difference` or `rand_tangent`. Are you on board with this? I'm keen not to document / export this functionality for now because we don't strictly need it until #97 is closer to completion. It's possible that we'll realise something here isn't quite right, and I would prefer not to force us to make multiple minor releases.

PR #96, opened 2020-07-07
Implement to_vec for Generator with identity

sethaxen:
This PR implements `to_vec` for generators with the identity function. It limits to the identity function until we have a way to handle closures. This should enable FD to be used for testing of functions that take generic iterators, so long as only this simple iterator is used in the test.

### DIFF
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -107,3 +107,13 @@ function to_vec(d::Dict)
     end
     return d_vec, Dict_from_vec
 end
+
+# Enable basic conversion of generators
+# just identity for now to avoid doing the wrong thing for closures
+function to_vec(g::Base.Generator{<:Any,typeof(identity)})
+    g_vec, back = to_vec(collect(g))
+    function Generator_from_vec(v)
+        return Base.Generator(identity, back(v))
+    end
+    return g_vec, Generator_from_vec
+end
--- a/test/to_vec.jl
+++ b/test/to_vec.jl
@@ -83,6 +83,12 @@ end
                 test_to_vec(Dict(:a=>3 + 2im, :b=>randn(T, 10, 11), :c=>(5+im, 2-im, 1+im)))
             end
         end
+
+        @testset "Generator(identity, x)" begin
+            test_to_vec((x for x in randn(T, 3)))
+            test_to_vec((x for x in randn(T, 3, 4)))
+            test_to_vec((x for x in randn(T, 3, 4, 5)))
+        end
     end
 
     @testset "FillVector" begin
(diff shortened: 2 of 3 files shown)


src/to_vec.jl
```diff
@@ -107,3 +107,13 @@ function to_vec(d::Dict)
     end
     return d_vec, Dict_from_vec
 end
+
+# Enable basic conversion of generators
+# just identity for now to avoid doing the wrong thing for closures
```
oxinabox, line 112: we can use `fieldcount` on the type of the function to detect if it is a closure.

oxinabox:
can you tell me more about the motivation?
Should we instead define a `TestingIterator` maybe?
Possibily even one with configurable `HasEltype` and `IteratorSize`?

sethaxen:
> can you tell me more about the motivation?

Sure, this was motivated by https://github.com/JuliaDiff/ChainRules.jl/pull/226, where I implemented `frule`s that take iterators and wanted to use `ChainRulesTestUtils.frule_test` to test the rules No non-array non-tuple iterators are currently usable with `to_vec`, hence this PR.

> Should we instead define a `TestingIterator` maybe?
> Possibily even one with configurable `HasEltype` and `IteratorSize`?

We certainly could. Although `Generator` is about as simple as it gets and works fine for this purpose. What would be the advantage of a custom iterator here?

src/to_vec.jl
sethaxen: That's a good point, so we could just raise an error if the `fieldcount` is nonzero for now.

oxinabox:
> We certainly could. Although Generator is about as simple as it gets and works fine for this purpose. What would be the advantage of a custom iterator here?

Primarily it is about being more explict in our tests that we are not testing `Generators` but just testing a generic iterator.
Also we might want to make it not have `HasSize` and `HasEltype`, so that it is the worst iterator possible.

sethaxen:
> Primarily it is about being more explict in our tests that we are not testing `Generators` but just testing a generic iterator.
> Also we might want to make it not have `HasSize` and `HasEltype`, so that it is the worst iterator possible.

Sounds fair. Since this is for testing, it's probably best to add the iterator and its `to_vec` method to ChainRulesTestUtils.

test/to_vec.jl
```diff
@@ -83,6 +83,18 @@ end
                 test_to_vec(Dict(:a=>3 + 2im, :b=>randn(T, 10, 11), :c=>(5+im, 2-im, 1+im)))
             end
         end
+
+        @testset "Generator(identity, x)" begin
```
willtebbutt, line 87: Would be good to include at least one test with an array that doesn't contain `Real`s.

Issue #98, opened 2020-07-11
Accuracy with Float32s is bad


willtebbutt:
@wesselb and I had a discussion about this a while ago, and I completely forgot to raise an issue about it. While `FiniteDifferences` accuracy for functions of `Float64`s is as you would expect, other types are a different matter. Of particular concern is `Float32` as it's almost certainty the next most used type.

IIRC the issue is an issue of defaults. Specifically [this one](https://github.com/JuliaDiff/FiniteDifferences.jl/blob/53bd1d1062f3cb5d6e2a184561d2d94c570de4e8/src/methods.jl#L225). `float` always yields a `Float64`, and `eps(::Float64)` is very different from `eps(Float32)`. This changes the step-size calculation [here](https://github.com/JuliaDiff/FiniteDifferences.jl/blob/53bd1d1062f3cb5d6e2a184561d2d94c570de4e8/src/methods.jl#L262).

If someone has time to look into this it would be greatly appreciated. I'm not entirely sure what the correct solution is, but a good approach to solving it might be

1. recap how in principle to set the `eps` parameter that is causing the problems (perhaps @wesselb could help here?)
2. figure out how to set `eps` based on this and make the appropriate changes
3. test that it works

wesselb:
If I'm not mistaken, the parameter `eps` is used to calculate the round-off error and should be set to the machine epsilon of the corresponding data type. The issue is that `TINY` is `Float64`, so `bound` always is `Float64`, even if the function outputs `Float32`s. In that case, `eps` is incorrectly set to the machine epsilon of `Float64`s, so the step size will be off.

For the particular case of `Float64`s and `Float32`s, most likely changing `TINY` to `1f-20` will fix most of the issues.

willtebbutt:
Might make sense to cast it explicitly wherever it's used to the element type of the thing that it's being added to. This way if (for some reason) we ever use `Float16`s or some extended-precision stuff we're protected.

wesselb:
Yes, that's a good suggestion. We should carefully check the code, to ensure that the assumption of `Float64`s is not also subtly baked in in other places.

willtebbutt:
Actually, it might make sense just to make `TINY` called `add_tiny` or something to prevent accidentally forgetting to cast stuff.

wesselb:
Yes, that's an even better solution.

wesselb:
[...]

PR #99, opened 2020-07-13
add_tiny + tests

willtebbutt:
As discussed with @wesselb in #98 .

Wessel: I'm getting test failures with second-order methods and `Float32`s caused by a loss in accuracy. Any chance that you could take a look to see whether the accuracy is as expected, or whether there's something going on?

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -187,9 +187,7 @@ The recognized keywords are:
 * `adapt`: The number of adaptive steps to use improve the estimate of `bound`.
 * `bound`: Bound on the value of the function and its derivatives at `x`.
 * `condition`: The condition number. See [`DEFAULT_CONDITION`](@ref).
-* `factor`: Estimate of the relative error on the function evaluations a multiple of the
-    machine epsilon. Defaults to `1`.
-* `eps`: The assumed roundoff error. Defaults to `eps()` plus [`TINY`](@ref).
+* `eps`: The assumed roundoff error. Defaults to `add_tiny(eps())`.
 
 !!! warning
     Bounds can't be adaptively computed over nonstandard grids; passing a value for
@@ -224,7 +222,6 @@ function fdm(
     ::Val{true};
     condition=DEFAULT_CONDITION,
     bound=_estimate_bound(f(x), condition),
-    factor=1,
     eps=add_tiny(Base.eps(bound)),
     adapt=m.history.adapt,
     max_step=convert(T, 0.1),
@@ -262,7 +259,7 @@ function fdm(
     end
 
     # Set the step size by minimising an upper bound on the error of the estimate.
-    C₁ = eps * sum(abs, coefs) * factor
+    C₁ = eps * sum(abs, coefs)
     C₂ = bound * sum(n->abs(coefs[n] * grid[n]^p), eachindex(coefs)) / factorial(p)
     ĥ = min((q / (p - q) * C₁ / C₂)^(1 / p), max_step)
 
@@ -277,7 +274,7 @@ function fdm(
     m.history.step = ĥ
     m.history.accuracy = accuracy
 
-    return m, T.(dfdx)
+    return m, dfdx
 end
 
 # Handle inputs that aren't `AbstractFloat`s -- assume what you wanted was a `Float64`. This
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -10,11 +10,11 @@ derivatives.
 const DEFAULT_CONDITION = 100
 
 """
-    FiniteDifferences.TINY
+    add_tiny(x::Real)
 
-A tiny number added to some quantities to ensure that division by 0 does not occur.
+Add a tiny number, 10^{-20}, to `x` while preserving its type.
 """
-const TINY = 1e-20
+add_tiny(x::T) where {T<:Real} = x + T(1e-20)
 
 forward_grid(p::Int) = 0:(p - 1)
 backward_grid(p::Int) = (1 - p):0
@@ -62,7 +62,7 @@ where the keyword arguments can be any of
 * `adapt`: The number of adaptive steps to use improve the estimate of `bound`.
 * `bound`: Bound on the value of the function and its derivatives at `x`.
 * `condition`: The condition number. See [`DEFAULT_CONDITION`](@ref).
-* `eps`: The assumed roundoff error. Defaults to `eps()` plus [`TINY`](@ref).
+* `eps`: The assumed roundoff error. Defaults to `add_tiny(eps())`.
 """
 abstract type FiniteDifferenceMethod end
 
@@ -171,7 +171,7 @@ end
 
 
 # Estimate the bound on the function value and its derivatives at a point
-_estimate_bound(x, cond) = cond * maximum(abs, x) + TINY
+_estimate_bound(x, cond) = add_tiny(cond * maximum(abs, x))
 
 """
     fdm(m::FiniteDifferenceMethod, f, x[, Val(false)]; kwargs...) -> Real
@@ -218,14 +218,14 @@ julia> fdm(central_fdm(2, 1), exp, 0, Val(true))
 function fdm(
     m::M,
     f,
-    x,
+    x::T,
     ::Val{true};
     condition=DEFAULT_CONDITION,
     bound=_estimate_bound(f(x), condition),
-    eps=(Base.eps(float(bound)) + TINY),
+    eps=add_tiny(Base.eps(T(bound))),
     adapt=m.history.adapt,
-    max_step=0.1,
-) where M<:FiniteDifferenceMethod
+    max_step=T(0.1),
+) where {T<:AbstractFloat, M<:FiniteDifferenceMethod}
     if M <: Nonstandard && adapt > 0
         throw(ArgumentError("can't adaptively compute bounds over Nonstandard grids"))
     end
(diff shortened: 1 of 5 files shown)


src/methods.jl
```diff
@@ -10,11 +10,11 @@ derivatives.
...
 
 """
-    FiniteDifferences.TINY
+    add_tiny(x::Real)
 
-A tiny number added to some quantities to ensure that division by 0 does not occur.
+Add a tiny number, 10^{-20}, to `x` while preserving its type.
 """
-const TINY = 1e-20
+add_tiny(x::T) where {T<:Real} = x + T(1e-20)
```
oxinabox, line 17:
```suggestion
add_tiny(x::T) where {T<:Real} = x + convert(T, 1e-20)
```

more formally i think `convert` belongs here rather than a constructor.

src/methods.jl
```diff
@@ -218,14 +218,14 @@ julia> fdm(central_fdm(2, 1), exp, 0, Val(true))
...
     ::Val{true};
     condition=DEFAULT_CONDITION,
     bound=_estimate_bound(f(x), condition),
-    eps=(Base.eps(float(bound)) + TINY),
+    eps=add_tiny(Base.eps(T(bound))),
     adapt=m.history.adapt,
-    max_step=0.1,
-) where M<:FiniteDifferenceMethod
+    max_step=T(0.1),
+) where {T<:AbstractFloat, M<:FiniteDifferenceMethod}
```
oxinabox, line 228: Why not `Real`?
```suggestion
) where {T<:Real, M<:FiniteDifferenceMethod}
```

src/methods.jl
```diff
@@ -239,8 +239,8 @@ function fdm(
 
     p = m.p
     q = m.q
-    grid = m.grid
-    coefs = m.coefs
+    grid = T.(m.grid)
+    coefs = T.(m.coefs)
```
oxinabox, line 243: This will allocate.
Can we avoid allocating by moving the conversion into the `sum` ?
So it only is converting 1 element at a time, and thus can do it on the stack and not allocate?

src/methods.jl
willtebbutt: See line 280 (ish). If an `Integer` is provided because someone types `1` rather than `1.0` then we need control over what type it converts to. Longer term I think we probably want to disallow `Integer` inputs altogether, but that's not something to be addressed in this PR.

test/methods.jl
```diff
@@ -1,30 +1,37 @@
...
+        @testset "method-order=$order" for order in [1, 2, 3]
+            @test m(order, 0; bound=1)(foo.f, T(1)) isa T
+            @test m(order, 0; bound=1)(foo.f, T(1)) == T(foo.f(1))
+        end
+
+        @test m(10, 1; bound=1)(foo.f, T(1)) isa T
+        @test m(10, 1; bound=1)(foo.f, T(1)) ≈ T(foo.d1)
+
+        @test m(10, 2; bound=1)(foo.f, T(1)) isa T
+        @test m(10, 2; bound=1)(foo.f, T(1)) ≈ T(foo.d2)
```
wesselb, line 34: Hmm, this line fails for `Float32`s in combination with `forward_fdm` and `backward_fdm`, but it succeeds for `central_fdm`.

I removed the converts inside the `fdm` function, and that fixed it. Would we be happy with performing the step size and derivative estimation at the highest possible accuracy and only converting back to `T` when returning the answser?

There is one more issue with the tests, which is that `backward_fdm` will produce numbers outside the domain of `sqrt`. This can be fixed by using `(f=x -> sqrt(x + 1), d1=0.5 / sqrt(2), d2=-0.25 / 2^(3/2))` instead.

I've branched off and pushed the above proposed fixes to `wb/add-tiny-fixes`.

willtebbutt: Hmm operating at the highest accuracy available definitely sounds appealing.

> I removed the converts inside the fdm function, and that fixed it. Would we be happy with performing the step size and derivative estimation at the highest possible accuracy and only converting back to T when returning the answser?

I think that we want to be careful here. In particular, we probably want to make sure that we're actually taking the correct code path, in the sense that we should make sure we're actually evaluating `f` using `Float32`s if we're trying out `Float32`s.

I'm imagining a situation in which the `Float32` implementation of `f` has slightly different characteristics from the `Float64` implementation and, as such, finite-differencing the `Float64` implementation gives the "wrong" answer. I'm not sure that I have a good answer off the top of my head though.

I think I would generally like to think a bit about the consequences of evaluating the function at a different level of precision before we commit to it, specifically the difference between different floating-point precision levels as opposed to `Integer` vs float, which we know causes issues for e.g. exponentiation -- it might be that it's totally fine.

wesselb: > I'm imagining a situation in which the Float32 implementation of f has slightly different characteristics from the Float64 implementation and, as such, finite-differencing the Float64 implementation gives the "wrong" answer. I'm not sure that I have a good answer off the top of my head though.

Ah, yes, that's definitely something to get right. I might have overlooked that in my branch. Let me get that right and what happens.

At a first thought, I would be in favour of providing the same data type to the function as the point at which we are evaluating the derivative.

willtebbutt: Yeah, that's my thinking too. I can see nothing wrong with doing step-size calculations for operations involving low-precision at higher-precision.

wesselb: The tests then fail. Interesting. Do you have any thoughts on this? Perhaps using forward differences at `Float32` to estimate a second derivative is pushing the method too hard. I'm not sure.

willtebbutt: Yeah, maybe we just shouldn't be testing that 🤷‍♂️ . @oxinabox any thoughts?

oxinabox: I do not have any thoughts.
I will point out that the `widen` function exists. 
```
julia> widen(Float32)
Float64
```


Maybe @shashi  has opinions, @shashi has done work on finite differencing tools before

willtebbutt: In the absence of other immediate options, I've stopped testing the second-derivative stuff on `Float32`s.

src/methods.jl
```diff
@@ -261,13 +261,14 @@ function fdm(
...
-    ĥ = min((q / (p - q) * C₁ / C₂)^(1 / p), max_step)
+    ĥ = convert(T, min((q / (p - q) * C₁ / C₂)^(1 / p), max_step))
 
     # Estimate the accuracy of the method.
     accuracy = ĥ^(-q) * C₁ + ĥ^(p - q) * C₂
 
     # Estimate the value of the derivative.
-    dfdx = sum(i->coefs[i] * f(x + ĥ * grid[i]), eachindex(grid)) / ĥ^q
+    # @show typeof(coefs), typeof(ĥ^q)
+    dfdx = sum(i -> convert(T, coefs[i]) * f(T(x + ĥ * grid[i])), eachindex(grid)) / ĥ^q
```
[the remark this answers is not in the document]
willtebbutt: I wasn't under the impression that `x` could be an array? I thought we were assuming that it was also going to be a scalar, hence my additional constraint on the type at the top.

src/methods.jl
```diff
@@ -261,13 +261,14 @@ function fdm(
...
     C₂ = bound * sum(n->abs(coefs[n] * grid[n]^p), eachindex(coefs)) / factorial(p)
-    ĥ = min((q / (p - q) * C₁ / C₂)^(1 / p), max_step)
+    ĥ = convert(T, min((q / (p - q) * C₁ / C₂)^(1 / p), max_step))
 
     # Estimate the accuracy of the method.
     accuracy = ĥ^(-q) * C₁ + ĥ^(p - q) * C₂
 
     # Estimate the value of the derivative.
-    dfdx = sum(i->coefs[i] * f(x + ĥ * grid[i]), eachindex(grid)) / ĥ^q
+    # @show typeof(coefs), typeof(ĥ^q)
```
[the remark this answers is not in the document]
willtebbutt: Yes, yes it should.

src/methods.jl
```diff
@@ -261,13 +261,14 @@ function fdm(
     # Set the step size by minimising an upper bound on the error of the estimate.
     C₁ = eps * sum(abs, coefs)
     C₂ = bound * sum(n->abs(coefs[n] * grid[n]^p), eachindex(coefs)) / factorial(p)
-    ĥ = min((q / (p - q) * C₁ / C₂)^(1 / p), max_step)
+    ĥ = convert(T, min((q / (p - q) * C₁ / C₂)^(1 / p), max_step))
```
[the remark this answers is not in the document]
willtebbutt: I think so -- the convert on 271 doesn't affect `ĥ`, does it? Maybe I'm missing something.

src/methods.jl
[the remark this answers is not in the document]
wesselb: Yes, that makes total sense. My bad.

src/methods.jl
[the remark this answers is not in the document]
wesselb: Hmm, is the `T` inside `f` on line 271 then necessary? I was wondering whether we need both the `convert` here and the `T` inside `f` on line 271. I guess it is necessary in the case that `grid[i]` is not an integer, in which case `grid[i] * h\hat` can be of a different type.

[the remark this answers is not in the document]
willtebbutt: Yeah, I reckon so. And it's a zero-cost thing if it's the correct type anyway, so I'm not too worried.

Issue #101, opened 2020-07-20
Travis CI Documenter failure


mattBrzezinski:
This has been occurring a few times recently where documentation fails to build,

* https://travis-ci.org/github/JuliaDiff/FiniteDifferences.jl/jobs/709794230
* https://travis-ci.org/github/JuliaDiff/FiniteDifferences.jl/jobs/708490805
* https://travis-ci.org/github/JuliaDiff/FiniteDifferences.jl/jobs/707247012

```julia
julia --project=docs/ docs/make.jl
[ Info: SetupBuildDirectory: setting up build directory.
[ Info: Doctest: running doctests.
[ Info: ExpandTemplates: expanding markdown templates.
[ Info: CrossReferences: building cross-references.
[ Info: CheckDocument: running document checks.
┌ Warning: `curl -sI --proto =http,https,ftp,ftps 'https://codecov.io/github/JuliaDiff/FiniteDifferences.jl?branch=master' --max-time 10 -o /dev/null --write-out '%{http_code} %{url_effective} %{redirect_url}'` failed:
│   exception =
│    failed process: Process(`curl -sI --proto =http,https,ftp,ftps 'https://codecov.io/github/JuliaDiff/FiniteDifferences.jl?branch=master' --max-time 10 -o /dev/null --write-out '%{http_code} %{url_effective} %{redirect_url}'`, ProcessExited(28)) [28]
│    
└ @ Documenter.DocChecks ~/.julia/packages/Documenter/IGqbY/src/DocChecks.jl:205
[ Info: Populate: populating indices.
ERROR: LoadError: `makedocs` encountered an error. Terminating build
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] runner(::Type{Documenter.Builder.RenderDocument}, ::Documenter.Documents.Document) at /home/travis/.julia/packages/Documenter/IGqbY/src/Builder.jl:242
 [3] dispatch(::Type{Documenter.Builder.DocumentPipeline}, ::Documenter.Documents.Document) at /home/travis/.julia/packages/Documenter/IGqbY/src/Utilities/Selectors.jl:167
 [4] #3 at /home/travis/.julia/packages/Documenter/IGqbY/src/Documenter.jl:319 [inlined]
 [5] cd(::Documenter.var"#3#8"{Documenter.Documents.Document}, ::String) at ./file.jl:104
[...]

PR #103, opened 2020-07-22
update documenter version

oxinabox:
This does not solve #101 , no logic around how that is done has changed AFAIK.
But it does give us the option to increase the timeout, if that is what we think needs to be done.

Also it adds pretty slick dark mode

image

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -3,4 +3,4 @@ Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 
 [compat]
-Documenter = "~0.23"
+Documenter = "~0.25"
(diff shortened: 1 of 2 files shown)


docs/Project.toml
```diff
@@ -3,4 +3,4 @@ Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 
 [compat]
-Documenter = "~0.23"
+Documenter = "~0.25"
```
nickrobinson251, line 6: same thing
```suggestion
Documenter = "0.25"
```

mattBrzezinski:
Apart from NickR's comment, looks good to me! As mentioned in #101 I don't know if bumping the timeout will resolve the issue there, it may help but it seems like a bandaid if anything.

I'm really not sure what could be causing this, but it's strange.

PR #104, opened 2020-07-23
Tweak add_tiny

willtebbutt:
This came up in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/138

There's some functionality that we've got that returns a `Bool`. `add_tiny` doesn't like this, so this is a fix for that.

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -12,9 +12,11 @@ const DEFAULT_CONDITION = 100
 """
     add_tiny(x::Real)
 
-Add a tiny number, 10^{-20}, to `x` while preserving its type.
+Add a tiny number, 10^{-20}, to `x`, preserving the type. If `x` is an `Integer`, it is
+promoted to a `Float64`.
 """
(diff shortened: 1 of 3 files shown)


src/methods.jl
```diff
@@ -12,9 +12,11 @@ const DEFAULT_CONDITION = 100
 """
     add_tiny(x::Real)
 
-Add a tiny number, 10^{-20}, to `x` while preserving its type.
+Add a tiny number, 10^{-20}, to `x`, preserving the type. If `x` is an `Integer`, it is
+promoted to a `Float64`.
```
oxinabox, line 16: Not true:
```
julia> float(BigInt(1)) |> typeof
BigFloat
```

oxinabox:
```suggestion
Add a tiny number, 10^{-20}, to `x`, preserving the type. If `x` is an `Integer`, it is
promoted to a suitable floating point type.
```

willtebbutt:
Good catch. I've committed your suggestion -- I'll merge when tests pass.

Issue #107, opened 2020-08-29
use richardson extrapolation?


stevengj:
I wonder if it would be useful to add a method using [Richardson.jl](https://github.com/JuliaMath/Richardson.jl), which can automatically [extrapolate low-order finite differences to high-order accuracy](https://github.com/JuliaMath/Richardson.jl#numerical-derivatives).   Another nice thing about Richardson extrapolation is that it automatically determines an appropriate step size — you just give it a relatively large initial guess, and it extrapolates downward, stopping automatically when cancellation errors take over.

oxinabox:
@wesselb can you comment on this?

Also link to the PDF where all the current math is explained.

wesselb:
That's very interesting. @stevengj, if I understand correctly, both Richardson extrapolation and direct construction of higher-order finite differences account for a number of terms of the Taylor expansion of the function. Contrary to the direct construction, Richardson extrapolation proceeds iteratively starting from a low-order method, and has a way of automatically detecting convergence. Could you comment on the convergence criterion? Does Richardson extrapolation tend to be robust, even when the initial guess for the step size is bad?

I can imagine a combination of both packages for robust, high-precision derivatives: start with a moderate-order (?) finite difference method with an automatically determined (hopefully good) step size, and then iterate with Richardson extrapolation to further refine the estimate.

Moreover, combining the ideas from both approaches (a geometrically converging sequence of step sizes and automatic balancing of truncation and round-off error), we could consider "geometric grids" like this:

```julia
julia> method = FiniteDifferences.Nonstandard(vcat(-(3 .^ [3, 2, 1]), 3 .^ [0, 1, 2, 3]), 1)

julia> fdm(method, sin, 1) - cos(1)
-1.3322676295501878e-15
```

What are your thoughts?

stevengj:
> Does Richardson extrapolation tend to be robust, even when the initial guess for the step size is bad?

It's certainly much more robust than finite differences with a fixed step size.  The initial guess just needs to be large enough not to suffer catastrophic cancellation, but not so large that increasing the polynomial order does nothing for the fit.

For example, for differentiating `sin(x)` at `x=1` by extrapolating a first-order forward difference, it gets the correct answer to about 13 digits for initial step sizes `h` between about `h=0.01` and `h=6`, and gets 7 decimal places down to about `h=1e-8`.

[...]

PR #109, opened 2020-09-15
WIP: Rework and Simplify Internals

wesselb:
This PR simplifies the internals, which have become a little convoluted over time. Also adds `extrapolate_fdm` (#107).

Important changes:
- There is only one type `FiniteDifferenceMethod`, which implements an FDM on an arbitrary grid.
- There is no function `fdm` anymore. All estimates are executed by calling the FDM directly.
- Some keyword arguments have changed place.

TODO:
- [x] Update documentation.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -7,6 +7,7 @@ ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Richardson = "708f8203-808e-40c0-ba2d-98a6953ed40d"
 
 [compat]
 ChainRulesCore = "0.9"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.10.8"
+version = "0.11.0"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
@@ -12,6 +12,7 @@ Richardson = "708f8203-808e-40c0-ba2d-98a6953ed40d"
 [compat]
 ChainRulesCore = "0.9"
 julia = "1"
+Richardson = "1.2"
 
 [extras]
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
--- a/README.md
+++ b/README.md
@@ -24,38 +24,42 @@ Right now here are the differences:
  - FiniteDiff.jl is carefully optimized to minimize allocations
  - FiniteDiff.jl supports coloring vectors for efficient calculation of sparse Jacobians
 
-## Examples
+
+#### FDM.jl
+This package was formerly called FDM.jl. We recommend users of FDM.jl [update to FiniteDifferences.jl](https://github.com/JuliaDiff/FiniteDifferences.jl/issues/37).
+
+
+## Example: Scalar Derivatives
 
 Compute the first derivative of `sin` with a 5th order central method:
 
 ```julia
 julia> central_fdm(5, 1)(sin, 1) - cos(1)
--1.247890679678676e-13
+-2.4313884239290928e-14
 ```
+
 Compute the second derivative of `sin` with a 5th order central method:
 
 ```julia
 julia> central_fdm(5, 2)(sin, 1) + sin(1)
-9.747314066999024e-12
+-4.367495254342657e-11
 ```
 
-Construct a FiniteDifferences on a custom grid:
+The functions `forward_fdm` and `backward_fdm` can be used to construct
+forward differences and backward differences respectively.
+
+Alternatively, you can construct a finite difference method on a custom grid:
 
 ```julia
-julia> method, report = fdm([-2, 0, 5], 1, report=true)
-(FiniteDifferences.method, FiniteDifferencesReport:
+julia> method = FiniteDifferenceMethod([-2, 0, 5], 1)
+FiniteDifferenceMethod:
   order of method:       3
   order of derivative:   1
   grid:                  [-2, 0, 5]
-  coefficients:          [-0.357143, 0.3, 0.0571429]
-  roundoff error:        2.22e-16
-  bounds on derivatives: 1.00e+00
-  step size:             3.62e-06
-  accuracy:              6.57e-11
-)
+  coefficients:          [-0.35714285714285715, 0.3, 0.05714285714285714]
 
 julia> method(sin, 1) - cos(1)
--2.05648831297367e-11
+-8.423706177040913e-11
 ```
 
 Compute a directional derivative:
@@ -64,10 +68,78 @@ Compute a directional derivative:
 julia> f(x) = sum(x)
 f (generic function with 1 method)
 
-julia> central_fdm(5, 1)(ε -> f([1, 1, 1] + ε * [1, 2, 3]), 0) - 6
--2.922107000813412e-13
+julia> central_fdm(5, 1)(ε -> f([1, 1, 1] .+ ε .* [1, 2, 3]), 0) - 6
+-6.217248937900877e-15
 ```
 
-## FDM.jl
+## Example: Multivariate Derivatives
 
-This package was formerly called FDM.jl. We recommend users of FDM.jl [update to FiniteDifferences.jl](https://github.com/JuliaDiff/FiniteDifferences.jl/issues/37).
+Consider a quadratic function:
+
+```julia
+julia> a = randn(3, 3); a = a * a'
+3×3 Array{Float64,2}:
+  8.0663   -1.12965   1.68556
+ -1.12965   3.55005  -3.10405
+  1.68556  -3.10405   3.77251
+
+julia> f(x) = 0.5 * x' * a * x
+```
+
+Compute the gradient:
+
+```julia
+julia> grad(central_fdm(5, 1), f, x)[1] - a * x
+3-element Array{Float64,1}:
+ -1.2612133559741778e-12
+ -3.526068326209497e-13
+ -2.3305801732931286e-12
+```
+
+Compute the Jacobian:
+
+```julia
+julia> jacobian(central_fdm(5, 1), f, x)[1] - (a * x)'
+1×3 Array{Float64,2}:
+ -1.26121e-12  -3.52607e-13  -2.33058e-12
+```
+
+The Jacobian can also be computed for non-scalar functions:
+
+```julia
+julia> a = randn(3, 3)
+3×3 Array{Float64,2}:
+ -0.343783   1.5708     0.723289
+ -0.425706  -0.478881  -0.306881
+  1.27326   -0.171606   2.23671
+
+julia> f(x) = a * x
+
+julia> jacobian(central_fdm(5, 1), f, x)[1] - a
+3×3 Array{Float64,2}:
+ -2.81331e-13   2.77556e-13  1.28342e-13
+ -3.34732e-14  -6.05072e-15  6.05072e-15
+ -2.24709e-13   1.88821e-13  1.06581e-14
+```
+
+To compute Jacobian--vector products, use `jvp` and `j′vp`:
+ 
+```julia
+julia> v = randn(3)
+3-element Array{Float64,1}:
+ -1.290782164377614
+ -0.37701592844250903
+ -1.4288108966380777
+
+julia> jvp(central_fdm(5, 1), f, (x, v)) - a * v
+3-element Array{Float64,1}:
+ -1.3233858453531866e-13
+  9.547918011776346e-15
+  3.632649736573512e-13
+
+julia> j′vp(central_fdm(5, 1), f, x, v)[1] - a'x
+ 3-element Array{Float64,1}:
+  3.5704772471945034e-13
+  4.04121180963557e-13
+  1.2807532812075806e-12
+```
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -355,6 +355,23 @@ end
 
 _exponentiate_grid(grid::Vector, base::Int=3) = sign.(grid) .* base .^ abs.(grid) ./ base
 
+function _is_symmetric(vec::Vector; centre_zero=false, negate_half=false)
+    n = div(length(vec), 2)
+    half_sign = negate_half ? -1 : 1
+    if isodd(length(vec))
+        centre_zero && vec[n + 1] != 0 && return false
+        return vec[1:n] == half_sign .* reverse(vec[n + 2:end])
+    else
+        return vec[1:n] == half_sign .* reverse(vec[n + 1:end])
+    end
+end
+
+function _is_symmetric(m::FiniteDifferenceMethod)
+    grid_symmetric = _is_symmetric(m.grid, centre_zero=true, negate_half=true)
+    coefs_symmetric =_is_symmetric(m.coefs, negate_half=true)
+    return grid_symmetric && coefs_symmetric
+end
+
 """
     extrapolate_fdm(
         m::FiniteDifferenceMethod,
@@ -368,7 +385,8 @@ Use Richardson extrapolation to refine a finite difference method. This method u
 [`estimate_step`](@ref) to determine an appropriate initial step size for
 [`Richardson.extrapolate`](@ref).
 
-Takes further in keyword arguments for [`Richardson.extrapolate`](@ref).
+Takes further in keyword arguments for [`Richardson.extrapolate`](@ref). This method
+automatically sets `power = 2` if `m` is symmetric.
 
 # Arguments
 - `m::FiniteDifferenceMethod`: Finite difference method to estimate the step size for.
@@ -407,7 +425,8 @@ extrapolate_fdm(m::FiniteDifferenceMethod, f::Function, x::T; kw_args...) where 
 Use Richardson extrapolation to refine a finite difference method. This method requires
 a given initial step size for [`Richardson.extrapolate`](@ref).
 
-Takes further in keyword arguments for [`Richardson.extrapolate`](@ref).
+Takes further in keyword arguments for [`Richardson.extrapolate`](@ref). This method
+automatically sets `power = 2` if `m` is symmetric.
 
 # Arguments
 - `m::FiniteDifferenceMethod`: Finite difference method to estimate the step size for.
@@ -425,7 +444,12 @@ function extrapolate_fdm(
     h;
     kw_args...
 ) where T<:AbstractFloat
-    return extrapolate(h -> m(f, x, h), h; kw_args...)
+    if _is_symmetric(m)
+        power = 2
+    else
+        power = 1
+    end
+    return extrapolate(h -> m(f, x, h), h; power=power, kw_args...)
 end
 # Handle arguments that are not floats. Assume that converting to float is desired.
 extrapolate_fdm(m::FiniteDifferenceMethod, f::Function, x::T, h; kw_args...) where T<:Real =
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -365,7 +365,7 @@ _exponentiate_grid(grid::Vector, base::Int=3) = sign.(grid) .* base .^ abs.(grid
     ) where T<:AbstractFloat
 
 Use Richardson extrapolation to refine a finite difference method. This method uses
-[`estimate_step`](@ref) to determine an appropriate initialisation for
+[`estimate_step`](@ref) to determine an appropriate initial step size for
 [`Richardson.extrapolate`](@ref).
 
 Takes further in keyword arguments for [`Richardson.extrapolate`](@ref).
@@ -389,8 +389,44 @@ function extrapolate_fdm(
     kw_args...
 ) where T<:AbstractFloat
     h_conservative = estimate_step(m, f, x)[1] * factor
-    return extrapolate(h -> m(f, x, h), h_conservative; kw_args...)
+    return extrapolate_fdm(m, f, x, h_conservative; kw_args...)
 end
 # Handle arguments that are not floats. Assume that converting to float is desired.
 extrapolate_fdm(m::FiniteDifferenceMethod, f::Function, x::T; kw_args...) where T<:Real =
     extrapolate_fdm(m, f, float(x); kw_args...)
+
+"""
+    extrapolate_fdm(
+        m::FiniteDifferenceMethod,
+        f::Function,
+        x::T
+        h;
+        kw_args...
+    ) where T<:AbstractFloat
+
+Use Richardson extrapolation to refine a finite difference method. This method requires
+a given initial step size for [`Richardson.extrapolate`](@ref).
+
+Takes further in keyword arguments for [`Richardson.extrapolate`](@ref).
+
+# Arguments
+- `m::FiniteDifferenceMethod`: Finite difference method to estimate the step size for.
+- `f::Function`: Function to evaluate the derivative of.
+- `x::T`: Point to estimate the derivative at.
+- `h`: Initial step size.
+
+# Returns
+- Estimate of the derivative.
+"""
+function extrapolate_fdm(
+    m::FiniteDifferenceMethod,
+    f::Function,
+    x::T,
+    h;
+    kw_args...
+) where T<:AbstractFloat
+    return extrapolate(h -> m(f, x, h), h; kw_args...)
+end
+# Handle arguments that are not floats. Assume that converting to float is desired.
+extrapolate_fdm(m::FiniteDifferenceMethod, f::Function, x::T, h; kw_args...) where T<:Real =
+    extrapolate_fdm(m, f, float(x), h; kw_args...)
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -329,7 +329,7 @@ function _make_adaptive_bound_estimator(
     if adapt >= 1
         estimate_derivative =
             constructor(q + 1, q, adapt=adapt - 1, condition=condition; kw_args...)
-        return (f, x) -> maximum(abs.(estimate_derivative(f, x)))
+        return (f, x) -> maximum(abs, estimate_derivative(f, x))
     else
         return _make_default_bound_estimator(condition=condition)
     end
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -1,4 +1,4 @@
-export FiniteDifferenceMethod, fdm, backward_fdm, forward_fdm, central_fdm
+export FiniteDifferenceMethod, fdm, backward_fdm, forward_fdm, central_fdm, extrapolate_fdm
 
 """
     add_tiny(x::Real)
@@ -90,6 +90,9 @@ automatically determined step size.
     to force a more conservative step size.
 - `max_step=convert(T, 0.1)`: Maximum step size.
 
+# Returns
+- Estimate of the derivative.
+
 # Examples
 
 ```julia-repl
@@ -140,6 +143,9 @@ step size.
 - `x::T`: Input to estimate derivative at.
 - `h`: Step size.
 
+# Returns
+- Estimate of the derivative.
+
 # Examples
 
 ```julia-repl
@@ -168,7 +174,7 @@ function (m::FiniteDifferenceMethod)(
     ) / h^m.q
 end
 # Handle arguments that are not floats. Assume that converting to float is desired.
-(m::FiniteDifferenceMethod)(f::Function, x::T, h) where T<:Real =  m(f, float(x), h)
+(m::FiniteDifferenceMethod)(f::Function, x::T, h) where T<:Real = m(f, float(x), h)
 
 # Check the method and derivative orders for consistency.
 function _check_p_q(p::Integer, q::Integer)
@@ -348,3 +354,43 @@ function _central_grid(p::Int)
 end
 
 _exponentiate_grid(grid::Vector, base::Int=3) = sign.(grid) .* base .^ abs.(grid) ./ base
+
+"""
+    extrapolate_fdm(
+        m::FiniteDifferenceMethod,
+        f::Function,
+        x::T;
+        factor::Int=1000,
+        kw_args...
+    ) where T<:AbstractFloat
+
+Use Richardson extrapolation to refine a finite difference method. This method uses
+[`estimate_step`](@ref) to determine an appropriate initialisation for
+[`Richardson.extrapolate`](@ref).
+
+Takes further in keyword arguments for [`Richardson.extrapolate`](@ref).
+
+# Arguments
+- `m::FiniteDifferenceMethod`: Finite difference method to estimate the step size for.
+- `f::Function`: Function to evaluate the derivative of.
+- `x::T`: Point to estimate the derivative at.
+
+# Factor
+- `factor::Int=1000`: Factor to amplify the estimated step size by.
+
+# Returns
+- Estimate of the derivative.
+"""
+function extrapolate_fdm(
+    m::FiniteDifferenceMethod,
+    f::Function,
+    x::T;
+    factor::Int=1000,
+    kw_args...
+) where T<:AbstractFloat
+    h_conservative = estimate_step(m, f, x)[1] * factor
+    return extrapolate(h -> m(f, x, h), h_conservative; kw_args...)
+end
+# Handle arguments that are not floats. Assume that converting to float is desired.
+extrapolate_fdm(m::FiniteDifferenceMethod, f::Function, x::T; kw_args...) where T<:Real =
+    extrapolate_fdm(m, f, float(x); kw_args...)
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -254,7 +254,7 @@ function estimate_step(
     f_x = float(f(x))
 
     # Estimate the bound and round-off error.
-    ε = add_tiny(maximum(eps.(f_x))) * factor
+    ε = add_tiny(maximum(eps, f_x)) * factor
     M = add_tiny(m.bound_estimator(f, x))
 
     # Set the step size by minimising an upper bound on the error of the estimate.
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -6,7 +6,7 @@ export FiniteDifferenceMethod, fdm, backward_fdm, forward_fdm, central_fdm, extr
 Add a tiny number, 10^{-40}, to `x`, preserving the type. If `x` is an `Integer`, it is
 promoted to a suitable floating point type.
 """
-add_tiny(x::T) where T<:Real = x + convert(T, 1e-40)
+add_tiny(x::T) where T<:AbstractFloat = x + convert(T, 1e-40)
 add_tiny(x::Integer) = add_tiny(float(x))
 
 """
@@ -19,7 +19,7 @@ derivatives.
 const DEFAULT_CONDITION = 100
 
 """
-    FiniteDifferenceMethod{G<:AbstractVector, C<:AbstractVector}
+    FiniteDifferenceMethod{G<:AbstractVector, C<:AbstractVector, E<:Function}
 
 A finite difference method.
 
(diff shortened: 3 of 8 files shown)


src/methods.jl
```diff
@@ -10,318 +19,378 @@ derivatives.
...
+"""
+function extrapolate_fdm(
+    m::FiniteDifferenceMethod,
+    f::Function,
+    x::T;
+    factor::Int=1000,
+    kw_args...
+) where T<:AbstractFloat
+    h_conservative = estimate_step(m, f, x)[1] * factor
+    return extrapolate(h -> m(f, x, h), h_conservative; kw_args...)
```
stevengj, line 392: Note that if it is a symmetric difference method, so that it only has powers of h^2 in the Taylor expansion of the error term, then it is worthwhile passing `power=2` to `extrapolate`.   Is this information available as part of `m`?

src/methods.jl
```diff
@@ -10,318 +19,378 @@ derivatives.
...
+# Returns
+- Estimate of the derivative.
+"""
+function extrapolate_fdm(
+    m::FiniteDifferenceMethod,
+    f::Function,
+    x::T;
+    factor::Int=1000,
+    kw_args...
+) where T<:AbstractFloat
```
stevengj, line 390: Why doe `T` need to be floating-point?

src/methods.jl
stevengj, line 392: Could be beneficial to pass `power=2` to `extrapolate` if `m` is a symmetric method, so that only powers of h^2 are present in the Taylor series for the error.  Is this information available in `m`?

src/methods.jl
```diff
@@ -10,318 +19,378 @@ derivatives.
...
-    # therefore handle this edge case here.
-    m.q == 0 && return m, f(x)
-
-    p = m.p
+function estimate_step(
+    m::FiniteDifferenceMethod,
+    f::Function,
+    x::T;
+    factor::Int=1,
+    max_step=convert(T, 0.1)
```
stevengj, line 250: This is a pretty strong assumption about the units of `x`…  If you're going to make this assumption, then I would just pass `1` or `0.1` to `Richardson.extrapolate`.

wesselb: Perhaps there's a better way to go about this. Limiting the step size is mainly for functions with a non-differentiability, like `abs`. I agree that `0.1` is pretty arbitrary and depends on the scale of `x`, and maybe we can choose it differently, e.g. capping it by a multiple of `abs(f(x))`.

My main worry with passing a fixed value like `1` or `0.1` is that is may be too large. Example:

```julia
julia> extrapolate(1, rtol=0) do h
           @show h
           (inv(1e-3 + h) - inv(1e-3)) / h
       end
h = 1.0
h = 0.125
h = 0.015625
(-8927.580356151784, 7928.579357150786)

julia> extrapolate(0.1, rtol=0) do h
           @show h
           (inv(1e-3 + h) - inv(1e-3)) / h
       end
h = 0.1
h = 0.0125
h = 0.0015625
(-83241.65749908325, 73340.66740007335)

julia> extrapolate(0.01, rtol=0) do h
           @show h
           (inv(1e-3 + h) - inv(1e-3)) / h
       end
h = 0.01
h = 0.00125
h = 0.00015625
h = 1.953125e-5
h = 2.44140625e-6
h = 3.0517578125e-7
h = 3.814697265625e-8
h = 4.76837158203125e-9
h = 5.960464477539063e-10
(-1.000000000029157e6, 3.0059716664254665e-5)
```

With the current implementation,

```julia
julia> extrapolate_fdm(forward_fdm(2, 1), inv, 1e-3)
(-1.0000000000231842e6, 3.273598849773407e-6)
```

An option would be to set the default cap to `Inf` and have the user set it whenever it is required.

src/methods.jl
wesselb: Ah, yes, nice! It's not available explicitly, but we should be able to determine that. Let me just add that.

README.md
```diff
@@ -24,38 +24,42 @@ Right now here are the differences:
...
--1.247890679678676e-13
+-2.4313884239290928e-14
 ```
+
 Compute the second derivative of `sin` with a 5th order central method:
 
 ```julia
 julia> central_fdm(5, 2)(sin, 1) + sin(1)
-9.747314066999024e-12
+-4.367495254342657e-11
```
oxinabox, line 45: Should I be worried about this loss of accuracy?
I guess not because the first order got more accurate

src/methods.jl
```diff
@@ -10,318 +19,438 @@ derivatives.
...
+    will be weighted by.
+- `bound_estimator::Function`: A function that takes in the function and the evaluation
+    point and returns a bound on the magnitude of the `length(grid)`th derivative.
+"""
+struct FiniteDifferenceMethod{G<:AbstractVector, C<:AbstractVector}
+    grid::G
+    q::Int
+    coefs::C
+    bound_estimator::Function
 end
```
oxinabox, line 39: Lets make it possible to access the `bounds_estimator` field in a type stable way.

```suggestion
struct FiniteDifferenceMethod{G<:AbstractVector, C<:AbstractVector, E<:Function}
    grid::G
    q::Int
    coefs::C
    bound_estimator::E
end
```

src/methods.jl
```diff
@@ -10,318 +19,438 @@ derivatives.
...
+    max_step=convert(T, 0.1)
+) where T<:AbstractFloat
+    # The automatic step size calculation fails if `m.q == 0`, so handle that edge case.
+    m.q == 0 && return f(x)
+    h, _ = estimate_step(m, f, x, factor=factor, max_step=max_step)
+    return m(f, x, h)
 end
+# Handle arguments that are not floats. Assume that converting to float is desired.
+(m::FiniteDifferenceMethod)(f::Function, x::T; kw_args...) where T<:Real =
+    m(f, float(x); kw_args...)
```
oxinabox, line 129: Blue style saule to only use the shortform function if it fits on oneline (92 characters)
otherwise write `function ...`

src/methods.jl
```diff
@@ -10,318 +19,438 @@ derivatives.
...
+```
+"""
+function (m::FiniteDifferenceMethod)(
+    f::Function,
+    x::T;
+    factor::Int=1,
+    max_step=convert(T, 0.1)
+) where T<:AbstractFloat
+    # The automatic step size calculation fails if `m.q == 0`, so handle that edge case.
+    m.q == 0 && return f(x)
```
oxinabox, line 123:
```suggestion
    iszero(m.q) && return f(x)
```

oxinabox (commented):
I haven't finished reviewing but i like it so far.

Did we get rid of the history field?
That that was problemantic so i would be glad to rid of it.

Can we do some tests on number of allocations
I deally anything that doesn't allocate from a single call should not allocate from having FDM called on it

src/methods.jl
wesselb: Ah, gotcha. I'll fix that all over.

wesselb:
> Did we get rid of the history field?

Yeah, I deleted it entirely. Perhaps a bit of a radical change, but I'm not sure that this had a use anywhere.

> I deally anything that doesn't allocate from a single call should not allocate from having FDM called on it

Do you mean by this that everything FDM does has to be without any allocation?

README.md
wesselb: I'm not sure that this loss of accuracy is significant. The whole step size estimation procedure relies on a lot of assumptions and approximations, so a change in accuracy like this could easily be due to another, unrelated change.

oxinabox:
> Do you mean by this that everything FDM does has to be without any allocation?

Yeah, more or less.
I mean once you have constructor the `FiniteDifferenceMethod` object, with its grid of coefficients etc,
all that `jvp` and `j′vp` and `grad` and `jacobian` are doing is invoking finite differencing in a loop.
and all that that is doing is invoking the function in a loop and computing a linear combination of the return values.
I,e, some scaled subtractions of scalars.
Scaling and substraction of scalars doesn't allocate.

src/methods.jl
```diff
@@ -10,318 +19,453 @@ derivatives.
...
-    p = length(grid)
-    _check_p_q(p, q)
-    coefs = _coefs(grid, p, q)
-    hist = History(; adapt=adapt, kwargs...)
-    return Nonstandard{typeof(grid), typeof(coefs)}(Int(p), Int(q), grid, coefs, hist)
+function (m::FiniteDifferenceMethod)(
+    f::Function,
+    x::T,
+    h
+) where T<:AbstractFloat
```
oxinabox, line 171: This all fits on one line under the 92 caracter limit.
```suggestion
function (m::FiniteDifferenceMethod)(f::Function, x::T, h) where T<:AbstractFloat
```

src/methods.jl
```diff
@@ -10,318 +19,453 @@ derivatives.
...
+    ) / h^m.q
 end
+# Handle arguments that are not floats. Assume that converting to float is desired.
+(m::FiniteDifferenceMethod)(f::Function, x::T, h) where T<:Real = m(f, float(x), h)
 
-# Check the method and derivative orders for consistency
+# Check the method and derivative orders for consistency.
 function _check_p_q(p::Integer, q::Integer)
-    q >= 0 || throw(ArgumentError("order of derivative must be nonnegative"))
+    q >= 0 || throw(ArgumentError("order of derivative must be non-negative"))
```
oxinabox, line 182: We can take this chance to swap to `DomainError`s if we wanted 
```suggestion
    q >= 0 || throw(DomainError(q, "order of derivative (`q`) must be non-negative"))
```

src/methods.jl
```diff
@@ -10,318 +19,453 @@ derivatives.
...
     q < p || throw(ArgumentError("order of the method must be strictly greater than that " *
                                  "of the derivative"))
```
oxinabox, line 184:
```suggestion
    q < p || throw(DomainError(
        (q, p),
        "order of the method (q) must be strictly greater than that of the derivative (p)",
    ))
```

src/methods.jl
```diff
@@ -10,318 +19,453 @@ derivatives.
...
-    # Check whether the method can be computed. We require the factorial of the
-    # method order to be computable with regular `Int`s, but `factorial` will overflow
-    # after 20, so 20 is the largest we can allow.
+    # Check whether the method can be computed. We require the factorial of the method order
+    # to be computable with regular `Int`s, but `factorial` will after 20, so 20 is the
+    # largest we can allow.
     p > 20 && throw(ArgumentError("order of the method is too large to be computed"))
```
oxinabox, line 188:
```suggestion
    p > 20 && throw(DomainError(p, "order of the method (`p`) is too large to be computed"))
```

src/methods.jl
```diff
@@ -10,318 +19,453 @@ derivatives.
...
 
-const _COEFFS_CACHE = Dict{Tuple{AbstractVector{<:Real}, Integer, Integer}, Vector{Float64}}()
-function _coefs(grid::AbstractVector{<:Real}, p::Integer, q::Integer)
-    return get!(_COEFFS_CACHE, (grid, p, q)) do
-        # For high precision on the \ we use Rational, and to prevent overflows we use Int128
-        # At the end we go to Float64 for fast floating point math (rather than rational math)
+# Compute coefficients for the method and cache the result.
+function _coefs(grid::AbstractVector{<:Real}, q::Integer)
+    return get!(_COEFFS_CACHE, (grid, q)) do
+        p = length(grid)
```
oxinabox, line 197: That is a nice improvement to not have to remember `p`

wesselb: Yes, it seemed redundant to carry around both.

wesselb:
> Yeah, more or less.

Okay, great! That indeed sounds very reasonable. I'll setup some tests for this and see how many allocations there are.

src/methods.jl
```diff
@@ -10,318 +19,453 @@ derivatives.
...
         C = [Rational{Int128}(g^i) for i in 0:(p - 1), g in grid]
         x = zeros(Rational{Int128}, p)
         x[q + 1] = factorial(q)
         return Float64.(C \ x)
     end
 end
 
+# Estimate the bound on the derivative by amplifying the ∞-norm.
+function _make_default_bound_estimator(; condition::Int=DEFAULT_CONDITION)
+    return (f, x) -> condition * maximum(abs.(f(x)))
```
oxinabox, line 208: We can avoid allocating a matrix this way.
```suggestion
    return (f, x) -> condition * maximum(abs, f(x))
```

src/methods.jl
```diff
@@ -10,318 +19,453 @@ derivatives.
...
+end
 
-# Estimate the bound on the function value and its derivatives at a point
-_estimate_bound(x, cond) = add_tiny(cond * maximum(abs, x))
+function Base.show(io::IO, x::FiniteDifferenceMethod)
```
oxinabox, line 211: the 2 arg show should not return something that is not valid julia code.
Since it is what 1arg `repr` hits.

Instread we want:
```suggestion
function Base.show(io::IO, m::MIME"text/plain", t::Template)
```

I think.
Which just controls the plain text display used by the REPL etc

src/methods.jl
```diff
@@ -10,318 +19,453 @@ derivatives.
...
-            bound=bound,
-            max_step=max_step,
-            adapt=(adapt - 1),
-        )
-        bound = _estimate_bound(dfdx, condition)
-    end
+    f_x = float(f(x))
+
+    # Estimate the bound and round-off error.
+    ε = add_tiny(maximum(eps.(f_x))) * factor
```
oxinabox, line 257:
```suggestion
    ε = add_tiny(maximum(eps, f_x)) * factor
```

src/methods.jl
```diff
@@ -10,318 +19,453 @@ derivatives.
...
+                grid,
+                q,
+                coefs,
+                _make_adaptive_bound_estimator(
+                    $fdm_fun,
+                    p,
+                    adapt,
+                    condition,
+                    geom=geom
+                )
```
oxinabox, line 295: This just fits in the line limit

```suggestion
                _make_adaptive_bound_estimator($fdm_fun, p, adapt, condition, geom=geom),
```

src/methods.jl
```diff
@@ -10,318 +19,453 @@ derivatives.
...
+function _make_adaptive_bound_estimator(
+    constructor::Function,
+    q::Int,
+    adapt::Int,
+    condition::Int;
+    kw_args...
+)
+    if adapt >= 1
+        estimate_derivative =
+            constructor(q + 1, q, adapt=adapt - 1, condition=condition; kw_args...)
```
oxinabox, line 337: I think the preferred line breaking is:
```suggestion
```suggestion
        estimate_derivative = constructor(
            q + 1, q, adapt=adapt - 1, condition=condition; kw_args...
        )
```

Rather than breaking after the =

src/methods.jl
```diff
@@ -10,318 +19,453 @@ derivatives.
...
+        return (f, x) -> maximum(abs.(estimate_derivative(f, x)))
```
oxinabox, line 338:
```suggestion
        return (f, x) -> maximum(abs, estimate_derivative(f, x))
```

src/methods.jl
stevengj: > e.g. capping it by a multiple of abs(f(x)).

This is just as as bad if not worse — it assumes that `f(x)` and `x` have the same units.

src/methods.jl
```diff
@@ -10,318 +19,453 @@ derivatives.
...
+_exponentiate_grid(grid::Vector, base::Int=3) = sign.(grid) .* base .^ abs.(grid) ./ base
+
+function _is_symmetric(vec::Vector; centre_zero=false, negate_half=false)
+    n = div(length(vec), 2)
+    half_sign = negate_half ? -1 : 1
+    if isodd(length(vec))
+        centre_zero && vec[n + 1] != 0 && return false
+        return vec[1:n] == half_sign .* reverse(vec[n + 2:end])
+    else
+        return vec[1:n] == half_sign .* reverse(vec[n + 1:end])
```
oxinabox, line 365: How do you feel about 
```suggestion
function _is_symmetric(vec::Vector; centre_zero=false, negate_half=false)
    half_sign = negate_half ? -1 : 1
    if isodd(length(vec))
        centre_zero && vec[end÷2 + 1] != 0 && return false
        return vec[1:end÷2] == half_sign .* reverse(vec[2+end÷2:end])
    else
        return vec[1:end÷2] == half_sign .* reverse(vec[1+end÷2:end])
```

I am not sure how i feel about it

src/methods.jl
```diff
@@ -10,318 +19,453 @@ derivatives.
...
+- `factor::Int=1000`: Factor to amplify the estimated step size by.
+
+# Returns
+- `Tuple{Real, Real}`: Estimate of the derivative and error.
+"""
+function extrapolate_fdm(
+    m::FiniteDifferenceMethod,
+    f::Function,
+    x::T;
+    factor::Int=1000,
```
oxinabox, line 406: is ther a reason this has to be an integer?

src/methods.jl
```diff
@@ -10,318 +19,451 @@ derivatives.
...
+    max_step=convert(T, 0.1)
+) where T<:AbstractFloat
+    # The automatic step size calculation fails if `m.q == 0`, so handle that edge case.
+    iszero(m.q) && return f(x)
+    h, _ = estimate_step(m, f, x, factor=factor, max_step=max_step)
+    return m(f, x, h)
+end
+# Handle arguments that are not floats. Assume that converting to float is desired.
+function (m::FiniteDifferenceMethod)(f::Function, x::T; kw_args...) where T<:Real
+    return m(f, float(x); kw_args...)
```
stevengj, line 129: This promotion scheme won't work for e.g. `Unitful` types, where `float(x)` does not give an `AbstractFloat`.  You'll get an infinite dispatch loop leading to a `StackOverflow` exception.

Maybe instead, do something like
```
(m::FiniteDifferenceMethod)(f::Function, x; kw_args...) = _evalrule(m, f, float(x); kw_args...)
```
and then define `_evalrule` to do the actual work.

Note that having `_evalrule` accept any type of `x` will allow many things beyond Unitful arguents. For example, it could also be used to take derivatives for complex arguments — the same finite-difference rules should work perfectly well, assuming the function is analytic at `x`.

Probably also set `max_step = 0.1*oneunit(x)` in `_evalrule`, for a similar reason — `convert` will fail for a unitful type.

src/methods.jl
stevengj: Every number you choose will be wrong for some function, of course…

(The basic problem with `inv(x)` at `x=1e-3` is that the radius of convergence of the Taylor series is so small at that point, of course.  Hence `extrapolate` is stopping because it detects a lack of convergence.   It's possible that I could use a better stopping criterion that somehow distinguishes between a lack of convergence because of floating-point cancellation — where we really should stop — and a lack of convergence due to the Taylor series — where should continue decreasing the step.)

src/methods.jl
oxinabox: Possibly we could just change this convience method to only be applies to subtypes of `<:Base.BitInteger` or `Union{Int32, Int64}` or `Int`
It's really just a convience for folk who are being lazy, and are writing something with interger literals.
And we relax the main defintion to accept any `Real`. (then it would work fine on e.g. `Rational`s i think?)
and if people give somehting bad then they can get an error then.

src/methods.jl
```diff
@@ -10,318 +19,451 @@ derivatives.
...
+function _coefs(grid::AbstractVector{<:Real}, q::Integer)
+    return get!(_COEFFS_CACHE, (grid, q)) do
+        p = length(grid)
+        # For high precision on the `\`, we use `Rational`, and to prevent overflows we use
+        # `Int128`. At the end we go to `Float64` for fast floating point math, rather than
+        # rational math.
         C = [Rational{Int128}(g^i) for i in 0:(p - 1), g in grid]
         x = zeros(Rational{Int128}, p)
         x[q + 1] = factorial(q)
         return Float64.(C \ x)
```
oxinabox, line 202: Should it always be `Float64` or sould it be based on `eltype(grid)`?

src/methods.jl
wesselb: > Every number you choose will be wrong for some function, of course…

Of course.

An important use case of the package is to automatically check the implementation of a derivative by comparing it with a numerical estimate. Since there are many rules to check, you really want to not bother appropriately configuring a finite difference for every one of them. For such use cases, it's important that the initialisation is as robust as it can be.

Regarding `max_step`, perhaps the two best options are to default it to `Inf` or to something like `max(maximum(abs, x), 0.1 * oneunit(x))`, which scales with `x`. Would you be happy with any of these?

src/methods.jl
wesselb: Not really. It just has to be something that will be multiplied to the output of `eps(x)`. Should we remove the type?

src/methods.jl
wesselb: The grid can be integers, so `eltype(grid)` won't always work. It should be the floating point type for the step size computation. `Float64` seems like a reasonable compromise. (I actually pretty much copied this piece of the code from the current `methods.jl`.) We could even make this part of the `FiniteDifferenceMethod` type, to e.g. allow step size computations with `Float64`s or even `Rational`s even though the input to the function is e.g. `Float32`.

src/methods.jl
wesselb: Also happy with that. Should there be spaces around the operators inside the `[...]`s?

src/methods.jl
wesselb: Just repeating it here to remove line 330.

```suggestion
        estimate_derivative = constructor(
            q + 1, q, adapt=adapt - 1, condition=condition; kw_args...
        )
```

src/methods.jl
wesselb: Gotcha! Thanks.

I'll change it to

```suggestion
function Base.show(io::IO, m::MIME"text/plain", x::FiniteDifferenceMethod)
```

src/methods.jl
oxinabox: I am not sure, I would need to look at th style guide closely.

I am not sure if my code if more or less clear

src/methods.jl
stevengj: `oftype(x, 0.1 * max(abs(x), oneunit(real(x))))` seems as good as anything, I suppose.

src/methods.jl
stevengj: What's wrong with my suggestion, i.e. calling `float` on all arguments and then passing it to a duck-typed helper routine?

src/methods.jl
stevengj: You could use `float(eltype(grid))`... if the user is employing more than double precision, it seems a shame to discard it.

oxinabox: I guess lets leave it as Float64 for now.
(I wrote it so its not like it has to be right)

src/methods.jl
stevengj: `::Real`?

src/methods.jl
oxinabox: Nothing.
It's also an option.

We could also call float and error if that didn't change the type,
Like SpecialFunctions.jl does.
https://github.com/JuliaMath/SpecialFunctions.jl/blob/master/src/gamma.jl#L551

I think your solution is probably better though than that.

src/methods.jl
wesselb: I like `float(eltype(grid))`, as it would indeed allow the user to specify a higher precision. It does feel a bit strange, however, that the user then is required to specify the data type of internal computations by the data type of the grid. It seems more logical to have it be a type parameter of `FiniteDifferenceMethod`.

oxinabox: hmm because we cache it by the value of the grid this could lead to weirdness.
since `isequal(Float32.(grid), Float64.(grid))`

Lets leave it for a future PR?

stevengj:
I updated Richardson.jl to add a new `breaktol` parameter to give more control over convergence.  If you set `breaktol=Inf`, it makes the convergence much more robust to initial guesses that are too large (assuming that it eventually converges for small enough `h`).

I added an [example to the documentation](https://github.com/JuliaMath/Richardson.jl#handling-problematic-convergence) showing how this fixes the problem you pointed out above with the derivative of `1/x` for small `x` — you can now put in an arbitrarily large initial guess for that function and it will still converge (at the expense of more iterations, of course).

So, I would require Richardson.jl 1.2.0 and set `breaktol=Inf`.   (At worst, it should shrink the stepsize `h` until your finite-difference approximation returns `NaN` from 0/0, but it will still terminate and return the best error obtained prior to that point.  In most cases it will hit the `rtol` before that point, however.)

wesselb:
Nice! That sounds great; thanks for adding this. I'll default to `breaktol = 10`, which appears to work for a few of the more tricky cases. It is probably a good idea to setup more elaborate and rigorous tests.

I am away for another week, but will then get back to this PR to wrap it up.

stevengj:
`breaktol = Inf` should be fine as well — the iteration will still halt due to the other stopping criteria.

oxinabox:
@wesselb  how close are we getting to this being mergable?

oxinabox:
Allocation performance is already better than what it was before:

using `const _fdm = central_fdm(5, 1; adapt=5)`
### Current release:
```julia
julia> @btime $_fdm( sin, 0.2);
  9.843 μs (254 allocations: 6.81 KiB)
```
### this PR
```julia
julia> @btime $_fdm(sin, 0.2);
  4.767 μs (146 allocations: 4.23 KiB)
```

src/methods.jl
```diff
@@ -10,318 +19,446 @@ derivatives.
...
+- `Tuple{T, T}`: Estimated step size and an estimate of the error of the finite difference
+    estimate.
+"""
+function estimate_step(
+    m::FiniteDifferenceMethod,
+    f::Function,
+    x::T;
+    factor::Int=1,
+    max_step=convert(T, 0.1)
+) where T<:AbstractFloat
```
oxinabox, line 251:
```suggestion
)::Tuple{T, T} where T<:AbstractFloat
```
I did some quick investigation, and this is currently failing to type-infer.
Possibly its not type-stable.
But anyway adding this convert-constraint gives a 3x performance vs
before
https://github.com/JuliaDiff/FiniteDifferences.jl/pull/109#issuecomment-701342044


```
julia> @btime $_fdm(sin, 0.2);
  1.393 μs (34 allocations: 1.66 KiB)
```

oxinabox:
I spent an hour or so digging into the remaining allocations
beyond https://github.com/JuliaDiff/FiniteDifferences.jl/pull/109#discussion_r497451612

They are all in `estimate_step`,
I think they boild down to the fact that adaption causes it to construct a new finite differencing method each time.
I think the solution is that adaptive FiniteDifferenceMethods should, when first constructed, construct the adapted versions of themselves.
This could be done with a field that just stores the next adaption (which itself would have a field that store the next adaption).
Beyond the scope of this PR maybe?

src/methods.jl
wesselb: This would assumes that if the input of `f` is of type `T` then the output is also of type `T`. Is that an assumption we're willing to make?

oxinabox: Hmm probably not, but it is what the docstring says it does

src/methods.jl
```diff
@@ -10,318 +19,446 @@ derivatives.
...
+    ε = add_tiny(maximum(eps, f_x)) * factor
+    M = add_tiny(m.bound_estimator(f, x))
 
-!!! warning
-    Bounds can't be adaptively computed over nonstandard grids; passing a value for
-    `adapt` greater than 0 when `m::Nonstandard` results in an error.
+    # Set the step size by minimising an upper bound on the error of the estimate.
+    C₁ = ε * sum(abs, m.coefs)
+    C₂ = M * sum(n -> abs(m.coefs[n] * m.grid[n]^p), eachindex(m.coefs)) / factorial(p)
+    h = convert(T, min((q / (p - q) * C₁ / C₂)^(1 / p), max_step))
```
oxinabox, line 263: This change gives the same performance improvement as https://github.com/JuliaDiff/FiniteDifferences.jl/pull/109#discussion_r497451612
but doesn't say anything about the `accuracy` argument
```suggestion
    # type-inference fails on this, so we annotate it, which gives big performance benifits
    h::T = convert(T, min((q / (p - q) * C₁ / C₂)^(1 / p), max_step))
```

src/methods.jl
```diff
@@ -10,318 +19,446 @@ derivatives.
...
+    adapt::Int,
+    condition::Int;
+    kw_args...
+)
+    if adapt >= 1
+        estimate_derivative = constructor(
+            q + 1, q, adapt=adapt - 1, condition=condition; kw_args...
         )
-        bound = _estimate_bound(dfdx, condition)
+        return (f, x) -> maximum(abs, estimate_derivative(f, x))
```
oxinabox, line 333: `estimate_derivative` always returns a scalar so no need for `maximum` is there?
```suggestion
        return (f, x) -> abs(estimate_derivative(f, x))
```

src/methods.jl
```diff
@@ -10,318 +19,446 @@ derivatives.
...
+    p = length(m.coefs)
+    q = m.q
+    f_x = float(f(x))
 
-* `adapt`: The number of adaptive steps to use improve the estimate of `bound`.
-* `bound`: Bound on the value of the function and its derivatives at `x`.
-* `condition`: The condition number. See [`DEFAULT_CONDITION`](@ref).
-* `eps`: The assumed roundoff error. Defaults to `add_tiny(eps())`.
+    # Estimate the bound and round-off error.
+    ε = add_tiny(maximum(eps, f_x)) * factor
```
oxinabox, line 257: `f_x` must be a scalar so no need for `maximum` here

```suggestion
    ε = add_tiny(eps(f_x)) * factor
```

src/methods.jl
```diff
@@ -10,318 +19,446 @@ derivatives.
...
+-1.7741363933510002e-13
+```
 """
-function Nonstandard(grid::AbstractVector{<:Real}, q::Integer; adapt=0, kwargs...)
-    p = length(grid)
-    _check_p_q(p, q)
-    coefs = _coefs(grid, p, q)
-    hist = History(; adapt=adapt, kwargs...)
-    return Nonstandard{typeof(grid), typeof(coefs)}(Int(p), Int(q), grid, coefs, hist)
+function (m::FiniteDifferenceMethod)(f::Function, x::T, h) where T<:AbstractFloat
```
oxinabox, line 167: I found i could get a bit of a performance boost out of annoating each of these methods to be 
```suggestion
@inline function (m::FiniteDifferenceMethod)(f::Function, x::T, h) where T<:AbstractFloat
```

src/methods.jl
nickrobinson251: > spaces around the operators

style guide recommends to write this like e.g. `vec[(end ÷ 2 + 1):end])`

src/methods.jl
wesselb: Ah, I forgot that the methods support functions `R -> R^n`, so it does need to be `maximum(abs, ...)`.

src/methods.jl
wesselb: Yes, let's indeed leave it for a future PR.

src/methods.jl
wesselb: Thanks, @nickrobinson251.

wesselb:
@oxinabox and @stevengj, I've addressed all your comments. How does this now look?

src/methods.jl
```diff
@@ -1,4 +1,13 @@
...
+export FiniteDifferenceMethod, fdm, backward_fdm, forward_fdm, central_fdm, extrapolate_fdm
+
+"""
+    add_tiny(x::Real)
+
+Add a tiny number, 10^{-40}, to `x`, preserving the type. If `x` is an `Integer`, it is
+promoted to a suitable floating point type.
+"""
+add_tiny(x::T) where T<:AbstractFloat = x + convert(T, 1e-40)
+add_tiny(x::Integer) = add_tiny(float(x))
```
stevengj, line 10: For this to work with arbitrary real types, including dimensionful types, you should probably use something like
```jl
add_tiny(x::T) where T<:Number = x + oneunit(T) * oftype(float(real(one(T))), 1e-40)
```

However, philosophically I'm very suspicious of this — not only are you assuming an absolute scale for `x`, but also you are using `1e-40` independent of the precision.

Furthermore, in most cases this code will be equivalent to `float(x)`, since the `1e-40` will be rounded off.  What are you trying to accomplish here?

wesselb: Perhaps it's better to leave support for dimensionful types for a future PR. I'm not familiar with the `Unitful` package, so I'm not sure what other things would need to change. The typing should currently be restrictive enough to throw method errors if things are not converted to `AbstractFloat`s, so there should be no infinite loops like there was before.

The purpose of `add_tiny` is to add a number that is small in the absolute sense. An example where this is used is in the step size estimation. The absolute value on the roundoff error on `x` is computed by `eps`. If `x = 0.0`, then `eps(x)` is very small, and things can NaN out. To prevent this from happening, we use `add_tiny(eps(x))` instead.

Project.toml
```diff
@@ -7,6 +7,7 @@ ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Richardson = "708f8203-808e-40c0-ba2d-98a6953ed40d"
```
oxinabox, line 11: Need to bump the version.
Are we happy enough with the public API to call this v1?

Project.toml
```diff
@@ -7,6 +7,7 @@ ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
...
 
 [compat]
 ChainRulesCore = "0.9"
```
oxinabox, line 13: Need to specify compat for `Richardson`

src/methods.jl
oxinabox: Should we make some minimal change, if the above works we can do that,
and then later investigate unitful support more broadly.

src/methods.jl
oxinabox: It seems like you didn't resolve this one?

src/methods.jl
```diff
@@ -10,318 +19,394 @@ derivatives.
...
+    grid_fun = Symbol("_", direction, "_grid")
+    @eval begin function $fdm_fun(
+            p::Int,
+            q::Int;
+            adapt::Int=1,
+            condition::Real=DEFAULT_CONDITION,
+            geom::Bool=false
+        )
+            _check_p_q(p, q)
+            grid = collect($grid_fun(p))
```
oxinabox, line 288: Do we have to do this?
it creates a copy.
Can we just make all  grid functions alway return a Vectors?
(Possibly we should look into them returning StaticArrays, but beyond the scope of this PR)

src/methods.jl
oxinabox: Oh i see you other comment sayws we do need this

oxinabox (approved):
Basically LGTM
See last few comments.
Address those and should be good to go

Do we want deprecations?
What is this release going to be numbers as

src/methods.jl
wesselb: The above change would add for complex numbers a tiny real number to only the real part. I'm not sure what the behaviour for complex numbers should be, as tiny would then (most logically) refer to the modulus, and there are many complex numbers with modulus `1e-40`. I'm happy to go with the above change if you think that this is not important. Alternatively, we can restrict the scope of the function change the docstring of the function to only work on `AbstractFloat`s and `Integer`s; the function is only used internally anyway.

Project.toml
wesselb: I'm not sure what's best. For now, I've changed the version to `0.11.0`. Let me know if you want to go for a major release instead.

src/methods.jl
wesselb: I've changed it so that all grid functions always return a `Vector`.

src/methods.jl
stevengj: > The purpose of add_tiny is to add a number that is small in the absolute sense.

There is no such number — there is no such thing as “small in the absolute sense”. Smallness is always relative to something else. You are implicitly assuming some absolute scale here, which cannot possibly hold for arbitrary functions. 

For example, if the function I’m differentiating is `exp10(x)` and I want the derivative at `x=-50`, then `1e-40` is not a small number in comparison to my function value. Or if I have `log10(x)` and I’m calculating the derivative at `x=1e-50`, then `1e-40` is not a small number in comparison to `x`. 

The whole point of floating-point arithmetic is that it is scale-invariant (until you reach underflow/overflow). I suspect that you are breaking that here.

oxinabox: `add_tiny` has now been removed as of https://github.com/JuliaDiff/FiniteDifferences.jl/pull/125

PR #111, opened 2020-10-12
Defined to_vec for ChainRulesCore differential types

oxinabox:
We already had ChainRulesCore as a dependency.
ChainRulesTestUtils was type-piratically defining `to_vec` on `Composite{<:Tuple, <:Tuple}`
This PR adds for `AbstractZero` (useful if working with a deep structure containing some Zero parts).
and it adds for `Composite` on structs.

Supporting to_vec on Composites of structs will close the last parts of
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/24
which really should have been on this repo.
I don't think it alone is enough to do finite differencing on structs, but it is a first step.

This works without https://github.com/JuliaDiff/ChainRulesCore.jl/pull/230
but for the tests to pass it needs that, since equality was defined wong

### DIFF
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -139,3 +139,25 @@ function to_vec(d::Dict)
     end
     return d_vec, Dict_from_vec
 end
+
+
+# ChainRulesCore Differentials
+function FiniteDifferences.to_vec(x::Composite{P, T}) where{P, T<:Tuple}
+    x_tuple = convert(Tuple, x)
+    x_vec, back_tuple = FiniteDifferences.to_vec(x_tuple)
+    function CompositeTuple_from_vec(y_vec)
+        y_tuple = back_tuple(y_vec)
+        return Composite{P, typeof(y_tuple)}(y_tuple)
+    end
+    return x_vec, CompositeTuple_from_vec
+end
+
+
+function FiniteDifferences.to_vec(x::AbstractZero)
+    function from_vec_AbstractZero(z)
+        length(z) == 1  || throw(DimensionMismatch("tried to go back to $x from $z"))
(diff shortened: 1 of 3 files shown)


src/to_vec.jl
```diff
@@ -139,3 +139,26 @@ function to_vec(d::Dict)
...
+end
+
+
+function FiniteDifferences.to_vec(x::AbstractZero)
+    function from_vec_AbstractZero(z)
+        length(z) == 1  || throw(DimensionMismatch("tried to go back to $x from $z"))
+        iszero(first(z)) || throw(DomainError(first(z)))
+        return x
+    end
+    return [false], from_vec_AbstractZero
```
nickrobinson251, line 163:
```suggestion
    return [false], AbstractZero_from_vec
```

nickrobinson251 (approved):
Presuming the test failure is fixed by https://github.com/JuliaDiff/ChainRulesCore.jl/pull/230

src/to_vec.jl
```diff
@@ -139,3 +139,26 @@ function to_vec(d::Dict)
...
+    function Composite_from_vec(y_vec)
+        y_back = back_inner(y_vec)
+        return Composite{P, typeof(y_back)}(y_back)
+    end
+    return x_vec, Composite_from_vec
+end
+
+
+function FiniteDifferences.to_vec(x::AbstractZero)
+    function from_vec_AbstractZero(z)
```
oxinabox, line 158:
```suggestion
    function AbstractZero_from_vec(z)
```

willtebbutt (approved):
LGTM. Definitely good to have this done.

PR #115, opened 2020-11-20
Tweak Zero to vec

willtebbutt:
`to_vec` for `AbstractZero`s was returning a non-empty vector, which is both unnecessary and potentially incorrect (`AbstractZero`s are structural, so if `FiniteDifferences` were to perturb an `AbstractZero`, bad things would presumably happen).


mzgubic (approved):
LGTM. I wonder whether the `test_to_vec` function could be improved somehow?

PR #119, opened 2020-11-23
Run CI and deploy docs with GitHub Actions

sethaxen:
Basically the same as https://github.com/JuliaDiff/ChainRules.jl/pull/309, but with Documenter.

### DIFF
--- a/.github/workflows/Nightly.yml
+++ b/.github/workflows/Nightly.yml
@@ -22,7 +22,6 @@ jobs:
           - macOS-latest
           - windows-latest
         arch:
-          - x86
           - x64
         # 32-bit Julia binaries are not available on macOS
         exclude:
--- a/.github/workflows/Nightly.yml
+++ b/.github/workflows/Nightly.yml
@@ -1,0 +1,48 @@
+# CI for Julia nightly, separate workflow to avoid failing CI badge on nightly fail
+name: Nightly
+on:
+  push:
+    branches: [master]
+    tags: [v*]
+  pull_request:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  test:
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - 'nightly'
+        os:
+          - ubuntu-latest
+          - macOS-latest
+          - windows-latest
+        arch:
+          - x86
+          - x64
+        # 32-bit Julia binaries are not available on macOS
+        exclude:
+          - os: macOS-latest
+            arch: x86
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: ${{ matrix.version }}
+          arch: ${{ matrix.arch }}
+      - uses: actions/cache@v1
+        env:
+          cache-name: cache-artifacts
+        with:
+          path: ~/.julia/artifacts
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
+            ${{ runner.os }}-
+      - uses: julia-actions/julia-buildpkg@v1
+      - uses: julia-actions/julia-runtest@v1
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -22,7 +22,6 @@ jobs:
           - macOS-latest
           - windows-latest
         arch:
-          - x86
           - x64
         # 32-bit Julia binaries are not available on macOS
         exclude:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,52 @@
+name: CI
+on:
+  push:
+    branches: [master]
+    tags: [v*]
+  pull_request:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  test:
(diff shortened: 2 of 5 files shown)


sethaxen:
All x86 and/or nightly builds fail, but all 64-bit and release versions pass.

willtebbutt:
Hmm do we want to bother running on 32bit, or is 64bit sufficient?

sethaxen:
> Hmm do we want to bother running on 32bit, or is 64bit sufficient?

64 bit is sufficient with what we were doing with Travis. However, it was recommended to test ChainRules with 32-bit, and since the test suite depends so strongly on FiniteDifferences, it would be nice to make sure that defaults, etc are good for 32-bit here.

willtebbutt:
@sethaxen it doesn't look like we're currently running on 32bit builds, so I would suggest that we

1. open an issue about 32 bit to decide it we actually want to do it, and
2. only use 64 bit builds in this PR so that we can get CI back up and running in the mean time.

Does this sound reasonable to you?

sethaxen:
@willtebbutt okay done!

willtebbutt:
@oxinabox @mattBrzezinski what are your thoughts about the CI configuration? Would be good if one of you could take a look.

mattBrzezinski:
> @oxinabox @mattBrzezinski what are your thoughts about the CI configuration? Would be good if one of you could take a look.

I think this line will allow for nightly failures to not matter? (I'm not 100% familiar with GHA though)

> https://github.com/JuliaCloud/AWSS3.jl/blob/master/.github/workflows/CI.yml#L14

Then we can just add the `nightly` to the `CI.yml` file, and scrape the other one. Also, would be nice to test on `1.3`, I think that we use this package at Invenia. Someone correct me otherwise.

sethaxen:
> Then we can just add the `nightly` to the `CI.yml` file, and scrape the other one. Also, would be nice to test on `1.3`, I think that we use this package at Invenia. Someone correct me otherwise.

Unless GitHub actions has recently changed something, when you use `continue_on_error`, then it marks the workflow as passing even if it failed. So if any nightly build fails, it will still get a green check mark in the PR, and you'll need to actually open the logs to find out that anything failed. But if we don't use it, then if nightly fails, the whole CI is marked as failure in the badge. So I always have a separate CI for nightly so I can see actual failures without affecting the badge.

mattBrzezinski:
> > Then we can just add the `nightly` to the `CI.yml` file, and scrape the other one. Also, would be nice to test on `1.3`, I think that we use this package at Invenia. Someone correct me otherwise.
> 
> Unless GitHub actions has recently changed something, when you use `continue_on_error`, then it marks the workflow as passing even if it failed. So if any nightly build fails, it will still get a green check mark in the PR, and you'll need to actually open the logs to find out that anything failed. But if we don't use it, then if nightly fails, the whole CI is marked as failure in the badge. So I always have a separate CI for nightly so I can see actual failures without affecting the badge.

Ah I see, different use cases I suppose. In that case, do you know if it's possible for actions to inherit from another one? I would be nice to centralize the common code from `CI.yml` and `Nightly.yml` into one file, then have each of these inherit from them.

If not possible, this LGTM.

sethaxen:
> In that case, do you know if it's possible for actions to inherit from another one? I would be nice to centralize the common code from `CI.yml` and `Nightly.yml` into one file, then have each of these inherit from them.

I don't think so. GA doesn't have support for workflow templates like e.g. Azure does.

mattBrzezinski:
> > In that case, do you know if it's possible for actions to inherit from another one? I would be nice to centralize the common code from `CI.yml` and `Nightly.yml` into one file, then have each of these inherit from them.
> 
> I don't think so. GA doesn't have support for workflow templates like e.g. Azure does.

In that case, send it! (Unless Lyndon has anything to add)

mattBrzezinski (approved):

.github/workflows/Nightly.yml
```diff
@@ -0,0 +1,47 @@
+# CI for Julia nightly, separate workflow to avoid failing CI badge on nightly fail
+name: Nightly
```
nickrobinson251, line 2: oh, there's an unfortunate naming clash here.

This file is for the Nightly Version of Julia (as you say in the comment above). 
But we will also want an action that runs CI on master each night, and which notifies folks at Invenia if it fails... and unfortunately, we _also_ call that "nightly" (despite being unrelated to the Nightly Version of Julia). This is the thing to do with having cron jobs and email notification that came up before at https://github.com/JuliaDiff/ChainRules.jl/pull/309#issuecomment-730429184}

Where the unfortunateness compounds is that we probably will end up with another file for this "run the CI each night and notify us of failures" thing... which we'll probably want to call "Nightly" 🤦‍♂️ And the pain is real: here's a PR that adds a "Nightly.yml" file with this second, different meaning to the "Nightly.yml" here
https://github.com/invenia/Intervals.jl/pull/151/files

The good news is the is purely a naming clash... so let's just rename this file! 
Suggestions welcome. Clearly, naming things is difficult.

Also, sorry.

nickrobinson251 (commented):
Everything substantive looks good! But there's a small naming issue 😆 😭

.github/workflows/Nightly.yml
mattBrzezinski: #whatsinaname

nickrobinson251: Maybe just tweak to JuliaNightly.yml ?

sethaxen: To add to the complication, we can't have CI in the workflow name, because at some point in the past (maybe still), that would cause its status to be reflected in the CI status badge. I renamed to JuliaNightly.

nickrobinson251: thanks! Ooof this GitHub Actions stuff is fiddly. Thank you so much for tackling it!

sethaxen: No problem!

PR #121, opened 2020-11-24
Reimplement to_vec for Symmetric/Hermitian

sethaxen:
The current implementation of `to_vec` for `Symmetric`/`Hermitian` doesn't seem quite right. In particular, this implementation makes `j′vp` not compose. See the newly added test for example (this fails on the current version but not this PR).

This is why for testing `rrule`s for functions that take `Symmetric`/`Hermitian` as input in ChainRules, we have not been able to use ChainRulesTestUtils. With this change, that should no longer be the case. Note that this is a breaking change.

### DIFF
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -68,10 +68,11 @@ function to_vec(x::T) where {T<:LinearAlgebra.AbstractTriangular}
 end
 
 function to_vec(x::T) where {T<:LinearAlgebra.HermOrSym}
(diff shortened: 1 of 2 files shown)


willtebbutt:
Thanks for this PR. Will take a look tomorrow evening.

sethaxen:
After some testing, while this does solve the composition issue, it doesn't enable testing ChainRules' `rrule`s with FiniteDifferences. Not entirely sure why, but this particular vectorization doesn't get us there. I'll open a new PR if I work out a better one.

PR #123, opened 2020-11-30
Switch CI to GitHub Actions and run with 32-bit Julia

### LINKED ISSUE #122: Enable CI on 32-bit builds
In #119, we temporarily turned on 32-bit builds, which all failed. It would be good to run CI on 32-bit builds to work out if this is an issue with tolerances in the test suite or in the source itself, since it would be handy to use this package to check AD rules on 32-bit machines as well.

giordano:
Fix #122.


giordano:
Woops, I saw #122 but not #119 :sweat_smile:

Issue #124, opened 2020-11-30
Accuracy is a bit brittle


giordano:
The [README](https://github.com/JuliaDiff/FiniteDifferences.jl/blob/673d6fa25c6eb7ac8a97d3e4ea8ca250219d3f88/README.md) of this package convinced me to [use `FiniteDifferences` instead of `Calculus` in `Measurements`](https://github.com/JuliaPhysics/Measurements.jl/pull/82) for cases where I need a numerical derivative (for functions without known derivative or automatic differentiation isn't possible, e.g. calling into C libraries).  However I found that accuracy of `FiniteDifferences` is quite brittle (sometimes strongly depends on the number of points with big even/odd jumps, see example below) and I need to use a largish number of grid points to achieve the same accuracy I have with `Calculus`, which in comparison is also much faster (and makes me question the change of the backend).  For example, the derivative of `cosc` in 0 is -(pi^2)/3, with Julia v1.5.3 I get:
```julia
julia> using Calculus, FiniteDifferences, BenchmarkTools

julia> -(pi ^ 2) / 3
-3.289868133696453

julia> (p -> FiniteDifferences.central_fdm(p, 1)(cosc, 0)).(2:10)
9-element Array{Float64,1}:
 -3.2575126788312536 # 2
  0.0                # 3
 -3.2894144933275746 # 4
 -3.9730287078618036 # 5 
 -3.289860720421136  # 6
 -3.2898376423854927 # 7
 -3.2898681297554884 # 8
 -3.289868469475128  # 9
 -3.2898681348743732 # 10

julia> Calculus.derivative(cosc, 0) 
-3.2898678494407765
```
To achieve an accuracy of the same order as `Calculus` I need 8 grid points, and this is the performance comparison:
```julia
julia> @btime FiniteDifferences.central_fdm(8, 1)($cosc, 0)
  2.340 μs (25 allocations: 992 bytes)
-3.2898681297554884

julia> @btime Calculus.derivative($cosc, 0)
  60.591 ns (0 allocations: 0 bytes)
-3.2898678494407765
```

Is this behaviour really expected?

nickrobinson251:
Doesn't sound expected to me. Would you mind trying with FiniteDifferences v0.10, to see if this changed between versions?

cc @wesselb

wesselb:
[...]

PR #125, opened 2020-11-30
WIP: Remove `add_tiny` and add `estimate_magitude`

wesselb:
The function `add_tiny` was a bit of a hack to deal with zeros. The new addition `estimate_magnitude` deals with this in a better way, at the cost of more function evaluations.

This should address #124:

```julia
julia> (p -> FiniteDifferences.central_fdm(p, 1)(cosc, 0)).(2:10) .+ (pi ^ 2) / 3
9-element Array{Float64,1}:
 -2.487109898532124
 -3.1522345644852123e-6
 -4.327381120106111e-9
 -8.283298491562618e-10
  1.3939516207983615e-11
  3.93636234718997e-11
  9.558132063602898e-12
 -2.1227464230832993e-13
  1.213251721310371e-12
```

The package can be pushed to estimate at high accuracies:

```julia
julia> FiniteDifferences.central_fdm(10, 1, adapt=3)(cosc, 0) + (pi ^ 2) / 3
-3.019806626980426e-14
```

This PR also adds a test that checks that above two cases.

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -10,10 +10,9 @@ estimate. This function deals with the case that `f(x) = 0`.
 @inline function estimate_magitude(f, x)
     M = float(maximum(abs, f(x)))
     M > 0 && (return M)
-    # `f(x) = 0`, but it may not be zero around `x`. We conclude that `x` is likely a
+    # Ouch, `f(x) = 0`. But it may not be zero around `x`. We conclude that `x` is likely a
     # pathological input for `f`. Perturb `x`. Assume that the pertubed value for `x` is
-    # highly unlikely also a pathological value for `f`, so use that as the magnitude for
-    # `f`.
+    # highly unlikely also a pathological value for `f`.
     Δ = 0.1 * max(abs(x), one(x))
     return float(maximum(abs, f(x + Δ)))
 end
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -18,6 +18,18 @@ function estimate_magitude(f, x::T) where T<:AbstractFloat
 end
 
 """
+    estimate_roundoff_error(f, x::T) where T<:AbstractFloat
+
+Estimate the round-off error of `f(x)`. This function deals with the case that `f(x) = 0`.
+"""
+function estimate_roundoff_error(f, x::T) where T<:AbstractFloat
+    # Estimate the round-off error. It can happen that the function is zero around `x`, in
+    # which case we cannot take `eps(f(x))`. Therefore, we assume a lower bound that is
+    # equal to `eps(T) / 1000`, which gives `f` four orders of magnitude wiggle room.
+    return max(eps(estimate_magitude(f, x)), eps(T) / 1000)
+end
+
+"""
     FiniteDifferences.DEFAULT_CONDITION
 
 The default [condition number](https://en.wikipedia.org/wiki/Condition_number) used when
@@ -265,10 +277,8 @@ function estimate_step(
     p = length(m.coefs)
     q = m.q
 
-    # Estimate the round-off error. It can happen that the function is zero around `x`, in
-    # which case we cannot take `eps(f(x))`. Therefore, we assume a lower bound that is
-    # equal to `eps(T) / 1000`, which gives `f` four orders of magnitude wiggle room.
-    ε = max(eps(estimate_magitude(f, x)), eps(T) / 1000) * factor
+    # Estimate the round-off error.
+    ε = estimate_roundoff_error(f, x) * factor
 
     # Estimate the bound on the derivatives.
     M = m.bound_estimator(f, x)
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -1,13 +1,22 @@
 export FiniteDifferenceMethod, fdm, backward_fdm, forward_fdm, central_fdm, extrapolate_fdm
 
 """
-    add_tiny(x::Union{AbstractFloat, Integer})
+    estimate_magitude(f, x)
 
-Add a tiny number, 10^{-40}, to a real floating point number `x`, preserving the type. If
-`x` is an `Integer`, it is promoted to a suitable floating point type.
+Estimate the magnitude of `f` in a neighbourhood of `x`, assuming that the outputs of `f`
+have a "typical" order of magnitude. The result should be interpreted as a very rough
+estimate. This function deals with the case that `f(x) = 0`.
 """
-add_tiny(x::T) where T<:AbstractFloat = x + convert(T, 1e-40)
-add_tiny(x::Integer) = add_tiny(float(x))
+@inline function estimate_magitude(f, x)
+    M = float(maximum(abs, f(x)))
+    M > 0 && (return M)
+    # `f(x) = 0`, but it may not be zero around `x`. We conclude that `x` is likely a
+    # pathological input for `f`. Perturb `x`. Assume that the pertubed value for `x` is
+    # highly unlikely also a pathological value for `f`, so use that as the magnitude for
+    # `f`.
+    Δ = 0.1 * max(abs(x), one(x))
+    return float(maximum(abs, f(x + Δ)))
+end
 
 """
     FiniteDifferences.DEFAULT_CONDITION
@@ -209,7 +218,7 @@ end
 
 # Estimate the bound on the derivative by amplifying the ∞-norm.
 function _make_default_bound_estimator(; condition::Real=DEFAULT_CONDITION)
-    default_bound_estimator(f, x) = condition * maximum(abs, f(x))
+    @inline default_bound_estimator(f, x) = condition * estimate_magitude(f, x)
     return default_bound_estimator
 end
 
@@ -256,17 +265,16 @@ function estimate_step(
 ) where T<:AbstractFloat
     p = length(m.coefs)
     q = m.q
-    f_x = float(f(x))
 
     # Estimate the bound and round-off error.
-    ε = add_tiny(maximum(eps, f_x)) * factor
-    M = add_tiny(m.bound_estimator(f, x))
+    ε = eps(estimate_magitude(f, x)) * factor
+    M = m.bound_estimator(f, x)
 
     # Set the step size by minimising an upper bound on the error of the estimate.
     C₁ = ε * sum(abs, m.coefs)
     C₂ = M * sum(n -> abs(m.coefs[n] * m.grid[n]^p), eachindex(m.coefs)) / factorial(p)
     # Type inference fails on this, so we annotate it, which gives big performance benefits.
-    h::T = convert(T, min((q / (p - q) * C₁ / C₂)^(1 / p), max_step))
+    h::T = convert(T, min((q / (p - q) * (C₁ / C₂))^(1 / p), max_step))
 
     # Estimate the accuracy of the method.
     accuracy = h^(-q) * C₁ + h^(p - q) * C₂
@@ -292,7 +300,7 @@ for direction in [:forward, :central, :backward]
                 grid,
                 q,
                 coefs,
-                _make_adaptive_bound_estimator($fdm_fun, p, adapt, condition, geom=geom),
+                _make_adaptive_bound_estimator($fdm_fun, p, q, adapt, condition, geom=geom),
             )
         end
 
@@ -327,6 +335,7 @@ end
 
 function _make_adaptive_bound_estimator(
     constructor::Function,
+    p::Int,
     q::Int,
     adapt::Int,
     condition::Int;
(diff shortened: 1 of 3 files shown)


src/methods.jl
```diff
@@ -1,13 +1,21 @@
...
 
-Add a tiny number, 10^{-40}, to a real floating point number `x`, preserving the type. If
-`x` is an `Integer`, it is promoted to a suitable floating point type.
+Estimate the magnitude of `f` in a neighbourhood of `x`, assuming that the outputs of `f`
+have a "typical" order of magnitude. The result should be interpreted as a very rough
+estimate. This function deals with the case that `f(x) = 0`.
 """
-add_tiny(x::T) where T<:AbstractFloat = x + convert(T, 1e-40)
-add_tiny(x::Integer) = add_tiny(float(x))
+@inline function estimate_magitude(f, x)
```
willtebbutt, line 10: Did you find that force-inlining was necessary here for some reason?

src/methods.jl
```diff
@@ -1,13 +1,21 @@
 export FiniteDifferenceMethod, fdm, backward_fdm, forward_fdm, central_fdm, extrapolate_fdm
 
 """
-    add_tiny(x::Union{AbstractFloat, Integer})
+    estimate_magitude(f, x)
 
-Add a tiny number, 10^{-40}, to a real floating point number `x`, preserving the type. If
-`x` is an `Integer`, it is promoted to a suitable floating point type.
+Estimate the magnitude of `f` in a neighbourhood of `x`, assuming that the outputs of `f`
```
willtebbutt, line 6: You've said a neighbourhood here and removed type constraints on `x` that previously ensured it was real. This suggests to me that you expect this new method to work for any type where there's some implicit assumptions around `x` being an element of some metric space, however, there aren't tests to that effect and it's not clear to me when you can / cannot expect `one` to be defined. So

1. is my assessment correct?
2. if so, could you add some tests?
3. if not, could you re-introduce the type-constraints and introduce tests for at least `Float64` and `Float32`?

src/methods.jl
```diff
@@ -1,13 +1,21 @@
...
 """
-add_tiny(x::T) where T<:AbstractFloat = x + convert(T, 1e-40)
-add_tiny(x::Integer) = add_tiny(float(x))
+@inline function estimate_magitude(f, x)
+    M = float(maximum(abs, f(x)))
+    M > 0 && (return M)
+    # Ouch, `f(x) = 0`. But it may not be zero around `x`. We conclude that `x` is likely a
+    # pathological input for `f`. Perturb `x`. Assume that the pertubed value for `x` is
+    # highly unlikely also a pathological value for `f`.
+    Δ = 0.1 * max(abs(x), one(x))
```
willtebbutt, line 16: Will e.g. promote to `Float64` from `Float32`?

src/methods.jl
```diff
@@ -209,7 +217,7 @@ end
 
 # Estimate the bound on the derivative by amplifying the ∞-norm.
 function _make_default_bound_estimator(; condition::Real=DEFAULT_CONDITION)
-    default_bound_estimator(f, x) = condition * maximum(abs, f(x))
+    @inline default_bound_estimator(f, x) = condition * estimate_magitude(f, x)
```
willtebbutt, line 220: Again, do we need to force-inline?

willtebbutt (changes requested):
Thanks for sorting this out promptly!

1. Would it make sense to construct some test cases where the conditions present in #124 are _nearly_ satisfied, for the sake of robustness testing? i.e. what if `M` is only very slightly greater than `0` or something?
2. Some unit testing for `estimate_magnitude` would be a good idea, particularly around input types that are not `Float64`s.
3. Integration testing for `Float32`s would be a good idea if you think we know how to construct them properly.

src/methods.jl
wesselb: Hmm, not necessarily. I'll remove the inlines (also your comment below).

src/methods.jl
wesselb: Hmm, no, the method is expected to only be run for `x`s that are `AbstractFloat`s. I'll add the type costraint and also add tests for `Float32` and `Float64`.

src/methods.jl
wesselb: Good catch. Added a `convert` and a test.

src/methods.jl
wesselb: Removed this.

wesselb:
To answer your questions,

1. I'm not sure. We could do that, but then we would have to determine an arbitrary threshold that determines what a pathological output of `f(x)` is. Admittedly, in perturbing `x`, we already have such an arbitrary threshold. I'm on the fence. What would you propose?

2. I've added type constraints and tests for `estimate_magnitude`.

3. Yeah, I fully agree that integration testing for `Float32`s is something that we should do properly, but my sense is that that is bigger than this PR. A whole test suite of hard cases for `Float64`s and `Float32`s and expected accuracies would be great.

src/methods.jl
willtebbutt: Great, thanks!

willtebbutt:
1. Hmmm good point. I think we're fine as-is (in this PR).
2. Thanks.
3. Fair point, let's leave it for the future.

willtebbutt (approved):
Approved subject to patch-version bump.

edit: we need CI back up and running. This cannot be merged until that happens.

wesselb:
Yes, ugh, that's annoying. Moreover, @oxinabox will check tomorrow if these fixes help with some other accuracy issues that he was encountering, so perhaps good to also wait for that.

willtebbutt:
In the mean time @giordano could you confirm if this branch resolves the issues that you've been encountering? In particular, are there any test cases over and above the one you provided in #124 that you've been having particular problems with?

giordano:
Ok, I can test this later today.  There were indeed a couple of other functions that were failing to achieve the same accuracy as `Calculus`, `cosc` was the first one.  Thanks!

giordano:
Another similar case, expected derivative in 0 is 0:
```julia
julia> Calculus.derivative(sinc, 0)
0.0

# v0.11.3
julia> (p -> FiniteDifferences.central_fdm(p, 1)(sinc, 0)).(2:10)
9-element Array{Float64,1}:
  0.0
  0.0
  5.444645613907266e-13
  2.7755575615628914e-16
 -9.538516174483089e-15
  2.42861286636753e-16
 -3.7404985543720125e-15
 -4.822111254688986e-16
  2.809173270009664e-15

# this PR
julia> (p -> FiniteDifferences.central_fdm(p, 1)(sinc, 0)).(2:10)
9-element Array{Float64,1}:
  0.0
  0.0
  5.444645613907266e-13
  4.312032877555794e-14
 -9.538516174483089e-15
  2.3967209127499836e-15
 -3.7404985543720125e-15
 -4.822111254688986e-16
  2.809173270009664e-15
```
I wouldn't say the accuracy in this case is bad, it's just that comparing with 0 is complicated.

Also the derivative of this function `f(t)` should be 0 (OK, I must admit that I have some very sick test cases)
```julia
julia> f(t) = (x -> QuadGK.quadgk(sin, -x, x)[1])(t)
f (generic function with 1 method)

julia> Calculus.derivative(f, 4)
7.065611032273702e-11

# v0.11.3
julia> (p -> FiniteDifferences.central_fdm(p, 1)(f, 4)).(2:10)
9-element Array{Float64,1}:
 -0.015243579281754268
  0.0
  4.131607608911521e-9
 -1.7704270075193033e-9
  2.7545394016191937e-9
 -8.188893009030013e-12
 -6.983545129351427e-11
 -3.0697586011503785e-11
 -2.164732008407795e-11

# this PR
julia> (p -> FiniteDifferences.central_fdm(p, 1)(f, 4)).(2:10)
9-element Array{Float64,1}:
  0.01641638262171357
 -1.5410326104441793e-5
  2.1935477946857812e-8
 -2.7022157015905038e-8
  9.047605793628396e-11
 -1.3604658271651211e-9
 -3.154690991530035e-11
  4.959996079211705e-11
  7.027698887063365e-13
```

giordano:
In the end I think I'll default to 8 grid points in `Measurements.jl` and call it a day

wesselb:
Ah, there was one other edge case that wasn't taken care of. The results look now as follows:

```julia
julia> (p -> FiniteDifferences.central_fdm(p, 1)(cosc, 0)).(2:10) .+ (pi ^ 2) / 3
9-element Array{Float64,1}:
 -2.487109898532124
 -3.1522345644852123e-6
 -4.327381120106111e-9
 -8.283298491562618e-10
  1.3939516207983615e-11
  3.93636234718997e-11
  9.558132063602898e-12
 -2.1227464230832993e-13
  1.213251721310371e-12

julia> (p -> FiniteDifferences.central_fdm(p, 1)(sinc, 0)).(2:10)
9-element Array{Float64,1}:
  0.0
  0.0
  5.444645613907266e-13
  4.312032877555794e-14
 -9.538516174483089e-15
  2.3967209127499836e-15
 -3.7404985543720125e-15
 -4.822111254688986e-16
  2.809173270009664e-15

julia> (p -> FiniteDifferences.central_fdm(p, 1)(f, 4)).(2:10)
9-element Array{Float64,1}:
 -2.3988103001624555e-13
  1.430803955495777e-12
 -2.4073864725037792e-14
 -2.2694686733052566e-15
  1.52180247233912e-13
 -1.1700430263506674e-15
  2.5255867408048164e-13
  7.01087210768824e-13
 -1.0381492304012043e-14
```

The latter two are near machine epsilon. I don't think you can get much better.

The hardest case is still `cosc`, but you can push it:

```julia
julia> FiniteDifferences.central_fdm(10, 1, adapt=3)(cosc, 0) + (pi ^ 2) / 3
-3.019806626980426e-14
```

You can also push Richardson extrapolation:

```julia
julia> FiniteDifferences.extrapolate_fdm(FiniteDifferences.central_fdm(2, 1), cosc, 0.0, contract=0.5)[1] + (pi ^ 2) / 3
-5.702105454474804e-13
```

giordano:
With the latest version all current tests pass already with `FiniteDifferences.central_fdm(4, 1)`, that's amazing!  I'll probably also play with the other extrapolation methods, but this is looking much better now, thanks!

wesselb:
Nice! I'm very happy to hear that. :)

src/methods.jl
```diff
@@ -1,13 +1,21 @@
...
-add_tiny(x::T) where T<:AbstractFloat = x + convert(T, 1e-40)
-add_tiny(x::Integer) = add_tiny(float(x))
+function estimate_magitude(f, x::T) where T<:AbstractFloat
+    M = float(maximum(abs, f(x)))
+    M > 0 && (return M)
+    # Ouch, `f(x) = 0`. But it may not be zero around `x`. We conclude that `x` is likely a
+    # pathological input for `f`. Perturb `x`. Assume that the pertubed value for `x` is
+    # highly unlikely also a pathological value for `f`.
+    Δ = convert(T, 0.1) * max(abs(x), one(x))
+    return float(maximum(abs, f(x + Δ)))
```
oxinabox, line 17: should we make this recursive?
```suggestion
    return estimate_magitude(f, x + Δ)
```

oxinabox: Though even recursive won't solve `sin(0.2*π*x)`

src/methods.jl
```diff
@@ -256,17 +264,20 @@ function estimate_step(
...
     q = m.q
-    f_x = float(f(x))
 
-    # Estimate the bound and round-off error.
-    ε = add_tiny(maximum(eps, f_x)) * factor
-    M = add_tiny(m.bound_estimator(f, x))
+    # Estimate the round-off error. It can happen that the function is zero around `x`, in
+    # which case we cannot take `eps(f(x))`. Therefore, we assume a lower bound that is
+    # equal to `eps(T) / 1000`, which gives `f` four orders of magnitude wiggle room.
+    ε = max(eps(estimate_magitude(f, x)), eps(T) / 1000) * factor
```
oxinabox, line 271: should we move this to a little function?
`estimate_roundoff_error`

oxinabox:
Yep, I can confirm that this fixes the problems we were having with ChainRules

oxinabox (approved):

src/methods.jl
wesselb: That would recurse infinitely for the function `x -> 0.0`. Moreover, a function might just actually be `0` in a neighbour around `x`.

src/methods.jl
wesselb: Totally! Pushed the change.

oxinabox:
I think this needs rebasing so CI will run.
Other than that this LGTM

wesselb:
@oxinabox The tests fail on nightly. Are we okay with that?

oxinabox:
> @oxinabox The tests fail on nightly. Are we okay with that?

yes this is fine.

PR #128, opened 2020-12-25
Implement to_vec for matrix factorizations

### LINKED ISSUE #114: Stackoverflow with QR
Using `j′vp` on a QR factorization result in a stackoverflow, like in the following example:

```julia
using FiniteDifferences
using LinearAlgebra

X = randn(5, 5)
j′vp(central_fdm(5, 1), X->qr(X).Q, randn(5, 5), X)
```

I believe it is due to the fact that `vec(Q)` return a `Base.ReshapedArray` (for `Q, R = qr(X)`).

Causes problem while testing https://github.com/JuliaDiff/ChainRules.jl/pull/306

Kolaru:
Fixes #114 and should help cleaning the tests in https://github.com/JuliaDiff/ChainRules.jl/pull/306.

### DIFF
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -140,6 +140,34 @@ function to_vec(X::T) where {T<:PermutedDimsArray}
     return x_vec, PermutedDimsArray_from_vec
 end
 
+# Factorizations
+
+function to_vec(x::SVD)
+    x_vec, back = to_vec([x.U, x.S, x.Vt])
+    function SVD_from_vec(v)
+        U, S, Vt = back(v)
+        return SVD(U, S, Vt)
+    end
+    return x_vec, SVD_from_vec
+end
+
+function to_vec(x::Cholesky)
+    x_vec, back = to_vec(x.factors)
+    function Cholesky_from_vec(v)
+        return Cholesky(back(v), x.uplo, x.info)
+    end
+    return x_vec, Cholesky_from_vec
+end
+
+function to_vec(x::S) where {S <: Union{LinearAlgebra.QRCompactWYQ, LinearAlgebra.QRCompactWY}}
+    x_vec, back = to_vec([x.factors, x.T])
+    function QRCompact_from_vec(v)
+        factors, T = back(v)
+        return S(factors, T)
+    end
+    return x_vec, QRCompact_from_vec
+end
+
 # Non-array data structures
 
 function to_vec(x::Tuple)
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -142,11 +142,14 @@ end
 
 # Factorizations
 
-function to_vec(x::SVD)
-    x_vec, back = to_vec([x.U, x.S, x.Vt])
+function to_vec(x::F) where {F <: SVD}
+    # Convert the vector S to a matrix so we can work with a vector of matrices
+    # only and inferrence work
+    v = [x.U, reshape(x.S, length(x.S), 1), x.Vt]
+    x_vec, back = to_vec(v)
     function SVD_from_vec(v)
-        U, S, Vt = back(v)
-        return SVD(U, S, Vt)
+        U, Smat, Vt = back(v)
+        return F(U, vec(Smat), Vt)
     end
     return x_vec, SVD_from_vec
 end
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -181,6 +181,7 @@ function to_vec(x::S) where {U, S <: Union{LinearAlgebra.QRCompactWYQ{U}, Linear
         factors, Tback = back(v)
         return S(factors, Tback)
     end
+
     return x_vec, QRCompact_from_vec
 end
 
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -162,11 +162,24 @@ function to_vec(x::Cholesky)
     return x_vec, Cholesky_from_vec
 end
 
-function to_vec(x::S) where {S <: Union{LinearAlgebra.QRCompactWYQ, LinearAlgebra.QRCompactWY}}
-    x_vec, back = to_vec([x.factors, x.T])
+function to_vec(x::S) where {U, S <: Union{LinearAlgebra.QRCompactWYQ{U}, LinearAlgebra.QRCompactWY{U}}}
+    # x.T is composed of upper triangular blocks. The subdiagonals elements
+    # of the blocks are abitrary. We make sure to set all of them to zero
+    # to avoid NaN.
+    blocksize, cols = size(x.T)
+    T = zeros(U, blocksize, cols)
+
+    for i in 0:div(cols - 1, blocksize)
+        used_cols = i * blocksize
+        n = min(blocksize, cols - used_cols)
+        T[1:n, (1:n) .+ used_cols] = UpperTriangular(view(x.T, 1:n, (1:n) .+ used_cols))
+    end
+
+    x_vec, back = to_vec([x.factors, T])
+
     function QRCompact_from_vec(v)
-        factors, T = back(v)
-        return S(factors, T)
+        factors, Tback = back(v)
+        return S(factors, Tback)
     end
     return x_vec, QRCompact_from_vec
 end
--- a/test/to_vec.jl
+++ b/test/to_vec.jl
@@ -136,9 +136,25 @@ end
                 @test all(!isnan, F_vec)
                 @inferred back(F_vec)
                 F_back = back(F_vec)
-                @test F.Q == F.Q
-                @test F.R == F.R
-                @test F_vec == first(to_vec(F))
+                @test F_back.Q == F.Q
+                @test F_back.R == F.R
+
+                # Make sure the result is consistent despite the arbitrary
+                # values in F.T.
+                @test first(to_vec(F)) == first(to_vec(F))
+
+                # Test F.Q as well since it has a special type. Since it is
+                # represented by the same T and factors matrices than F
+                # it needs the same special treatment.
+                Q = F.Q
+                @inferred to_vec(Q)
+                Q_vec, back = to_vec(Q)
+                @test Q_vec isa Vector
+                @test all(s -> s isa Real, Q_vec)
+                @test all(!isnan, Q_vec)
+                @inferred back(Q_vec)
+                Q_back = back(Q_vec)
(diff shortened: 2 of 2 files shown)


willtebbutt:
LGTM. Could you please bump the patch version?

willtebbutt:
I'm happy with this, but could someone else quickly take a look to make sure I've not missed anything obvious. @oxinabox @nickrobinson251 @wesselb ?

oxinabox:
I am on leave.
If noone has looked at it before the 4th of Jan, I will look at it that week.

test/to_vec.jl
```diff
@@ -93,6 +93,16 @@ end
             )
         end
 
+        @testset "Factorizations" begin
+            for dims in [(5, 5), (4, 6), (7, 3)]
+                M = randn(T, dims...)
+                P = M * M' + I  # Positive definite matrix
+                test_to_vec(svd(M); check_inferred = false)
```
wesselb, line 100: Is it an issue that type inference fails here?

wesselb (approved):
Looks good to me! Thanks for adding this. :) I've left two small comments.

src/to_vec.jl
```diff
@@ -110,6 +110,34 @@ function to_vec(X::T) where {T<:PermutedDimsArray}
...
+function to_vec(x::S) where {S <: Union{LinearAlgebra.QRCompactWYQ, LinearAlgebra.QRCompactWY}}
+    x_vec, back = to_vec([x.factors, x.T])
+    function QRCompact_from_vec(v)
+        factors, T = back(v)
+        return S(factors, T)
+    end
+    return x_vec, QRCompact_from_vec
+end
+
 # Non-array data structures
```
rkube, line 141: `x.T` consists of upper triangular blocks: https://github.com/JuliaLang/julia/pull/41363#discussion_r659375154
The lower part is undefined on purpose, NaNs can hide there. This line won't work:  
`x_vec, back = to_vec([x.factors, x.T])`

For application purposes it may be preferable to think of derivatives in `Q` and `R`, not in `factors` and `T`.
So `to_vec` would unroll `Q` and `R` into a matrix. To instantiate a new QR object from updated values, one would need to re-calculate the QR factorization:

```
function FiniteDifferences.to_vec(x::S) where {S <: Union{LinearAlgebra.QRCompactWYQ, LinearAlgebra.QRCompactWY}}
    x_vec, x_back = to_vec([x.Q x.R])
    function QRCompact_from_vec(v)
        Q_new, R_new = x_back(v)
        QR = S(Q_new * R_new)
        return S(QR.factors, QR.T)
    end
    return x_vec, QRCompact_from_vec
end
```

rkube:
What is the status of this PR? I see that some checks are failing but the logs have expired. Do you need help with this?

oxinabox:
> What is the status of this PR? 

It is waiting for @Kolaru 's to replay to Wessel's comments.
It seems it's dropped off their radar.

It is reasonable to make another PR to branching off there branch, then rebase and fix the tests, and the comments, and we would be able to merge it.
(They would still show up in the contributors since would keep there commit, if anyone is keeping score.)

Kolaru:
Indeed it kind of went off my radar. I think it may have been partially superseeded by #156. I will try to come back to it ASAP.

test/to_vec.jl
Kolaru: I fixed the inferrence.

src/to_vec.jl
Kolaru: I tried that but unfortunately recomputing the QR factorization is enough to introduce numerical errors, and make the `to_vec` test fail (since it test for exact equality).

As far as I understand the internal representation used by `to_vec` should not change the result and it should be fine. If it is not the case, I will need to be smarter, somehow.

Kolaru:
This should be ready now.

src/to_vec.jl
simeonschaub: > As far as I understand the internal representation used by to_vec should not change the result and it should be fine. If it is not the case, I will need to be smarter, somehow.

If it contains NaNs, it definitely does. This should be reusing the logic added in JuliaLang/julia#41363, in particular `_triuppers_qr`.

simeonschaub (changes requested):
The QRCompactWY case is still broken, see my comment above.

src/to_vec.jl
Kolaru: I am not sure to understand. The numerical instability I am referring to are those that prevent are introduced by recomputing the QR decomposition internally:

```julia
julia> X = rand(10, 10) ;

julia> F = qr(X) ;

julia> Xnew = F.Q * F.R ;

julia> Xnew == X
false

julia> Fnew = qr(Xnew) ;

julia> Fnew == F
false

julia> Fnew.Q == F.Q
false
```

The tests require `from_vec(to_vec(F)) == F` which is not guaranteed there.

To deal with NaNs in `T` is it preferrable to change the supported version of Julia and use `LinearAlgebra._triuppers_qr` or to reproduce here part of the logic?

simeonschaub: I just meant that by filtering out the undef values, we get both the correct behavior and avoid any numerical issues related to having to refactorize the recalculated matrix.

> To deal with NaNs in T is it preferrable to change the supported version of Julia and use LinearAlgebra._triuppers_qr or to reproduce here part of the logic?

It's probably easiest if we just copy that code from there for now. Would be good to mention in a comment that this is taken from Base though and that it can be removed once we drop support for older versions of Julia.

simeonschaub: BTW, a good test here would be to check that `to_vec(qr(A)) == to_vec(qr(A))`

Kolaru: I incorporated that, and added a test of NaN in test_to_vec along the way.

QR must unfortunaely be tested separately for now, because it compares the T matrices that have those arbitrary values.

src/to_vec.jl
```diff
@@ -162,11 +162,24 @@ function to_vec(x::Cholesky)
     return x_vec, Cholesky_from_vec
 end
 
-function to_vec(x::S) where {S <: Union{LinearAlgebra.QRCompactWYQ, LinearAlgebra.QRCompactWY}}
-    x_vec, back = to_vec([x.factors, x.T])
+function to_vec(x::S) where {U, S <: Union{LinearAlgebra.QRCompactWYQ{U}, LinearAlgebra.QRCompactWY{U}}}
```
simeonschaub, line 165: Are you sure this is valid for `QRCompactWYQ`? If it is, it should at least be tested.

src/to_vec.jl
```diff
@@ -162,11 +162,24 @@ function to_vec(x::Cholesky)
...
+        T[1:n, (1:n) .+ used_cols] = UpperTriangular(view(x.T, 1:n, (1:n) .+ used_cols))
+    end
+
+    x_vec, back = to_vec([x.factors, T])
+
     function QRCompact_from_vec(v)
-        factors, T = back(v)
-        return S(factors, T)
+        factors, Tback = back(v)
+        return S(factors, Tback)
```
simeonschaub, line 182: Note that using `S` like this here will not be type stable. Since it's not super performance critical, it might not be worth all the trouble though.

test/to_vec.jl
```diff
@@ -117,12 +118,27 @@ end
...
+                F = qr(M)
+                @inferred to_vec(F)
+                F_vec, back = to_vec(F)
+                @test F_vec isa Vector
+                @test all(s -> s isa Real, F_vec)
+                @test all(!isnan, F_vec)
+                @inferred back(F_vec)
+                F_back = back(F_vec)
+                @test F.Q == F.Q
+                @test F.R == F.R
```
simeonschaub, line 140: What does this test? Would be good to also add the `to_vec(qr(A)) == to_vec(qr(A))` test I proposed above.

test/to_vec.jl
```diff
@@ -117,12 +118,27 @@ end
         end
 
         @testset "Factorizations" begin
-            for dims in [(5, 5), (4, 6), (7, 3)]
+            # (100, 100) is needed to test for the NaNs that can appear in the
+            # qr(M).T matrix
+            for dims in [(5, 5), (4, 6), (7, 3), (100, 100)]
```
simeonschaub, line 123: I don't think it is necessary to test for that many different sizes. Just one size greater than 36x36 should be enough.

src/to_vec.jl
Kolaru: `QRCompactWYQ` is represented with the same `T` and `factors` matrices than `QRCompactWY`, yes. I added the tests.

src/to_vec.jl
Kolaru: Out of curiosity, why is that so?

test/to_vec.jl
Kolaru: This was an oversight from me, it should read `F_back.Q == F.Q`, to test that `back(to_vec(qr(A)))` correctly gives back the input.

Your suggested test was on the next line, I have reformulated it to make more explicit what is tested.

test/to_vec.jl
Kolaru: My idea was to test one all combinations of square-rectangular matrices for `Q` and `R`, just to be on the safe side.

src/to_vec.jl
simeonschaub: Perfect, thanks!

src/to_vec.jl
simeonschaub: See JuliaLang/julia#23618

test/to_vec.jl
simeonschaub: Maybe just:
```suggestion
            for dims in [(7, 3), (100, 100)]
```
? Those other two shouldn't hit any different paths, right? My only concern is that since we already test for different combinations of `T` here, testing lots of other combinations can quickly blow up the time these tests take with little improvement to actual coverage, if they all hit the same paths anyways. Not a huge deal right now, but as the test suite grows, that can become an issue, so I'd prefer to be more conservative here.

simeonschaub (approved):
One more comment, otherwise looks good to me. Thanks for pulling through!

Kolaru:
I integrated the last suggestion, that should indeed be enough to cover all cases.

simeonschaub:
Cool, thanks for the contribution!

PR #130, opened 2021-01-02
Improve Robustness and Performance

### LINKED ISSUE #129: gradient of non-smooth functions
Hi,
we are using FiniteDifferences in NNlib.jl to validate the automatic derivative compute through Zygote of some operators, like ReLU and maxpool,  that contain a few singularities. 
Since we test on a few random inputs, if FD evaluates the function in a small neighborhood of the input the FD estimate should be correct and not affected by the singularities with high probability. What we see instead is that sometimes the estimate is off.  
I couldn't figure out from the docs how to select a small grid spacing or something along the lines, I tried with
```julia
fdm = FiniteDifferenceMethod(1e-3 .* [-2, -1, 0, 1, 2], 1) 
```
but with no observable improvement (even worse, I get some cholesky factorization errors If I remember correctly).
Any help with this would be appreciated. 

Best,
Carlo

wesselb:
`master`:

```julia
julia> m = central_fdm(6, 1); @benchmark $m(sin, 1)
BenchmarkTools.Trial:
  memory estimate:  1.02 KiB
  allocs estimate:  56
  --------------
  minimum time:     2.220 μs (0.00% GC)
  median time:      2.307 μs (0.00% GC)
  mean time:        2.491 μs (1.43% GC)
  maximum time:     182.601 μs (98.30% GC)
  --------------
  samples:          10000
  evals/sample:     9
```

This PR:

```julia
julia> m = central_fdm(6, 1); @benchmark $m(sin, 1)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     451.449 ns (0.00% GC)
  median time:      466.278 ns (0.00% GC)
  mean time:        479.135 ns (0.00% GC)
  maximum time:     1.387 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     198
```

The current implementation is suboptimal in several ways:

1. The function `f` is evaluated at `x` and possibly again around `x` [to estimate the round-off error](https://github.com/JuliaDiff/FiniteDifferences.jl/blob/master/src/methods.jl#L29). This can be avoided by reusing the function evaluations of the bound estimator.

2. The bound estimator [may be run twice](https://github.com/JuliaDiff/FiniteDifferences.jl/blob/master/src/methods.jl#L361) if it returns zero identically, which requires even more function evaluations. This can be avoided by increasing the order of the bound estimator by one so that it can estimate the derivative in a neighbourhood of `x` instead of just at `x`.

3. The functions `estimate_magnitude` and `estimate_roundoff_error`, whilst fine in principle, feel a little ad hoc and complicate the logic.

This PR gets rid of `estimate_magnitude` and `estimate_roundoff_error` entirely by more cleverly using the function evaluations of the bound estimator (`_estimate_magnitudes`). Moreover, the PR makes sure that unnecessary allocations are avoided where possible.

Improvements:
- [x] Reduce number of allocations.
- [x] Reduce number of function evaluations and ensure that the number of function evaluations is always the same.
- [x] Improve robustness of adaptation.
- [x] Improve mechanism for step size capping.
- [x] Eliminate dependence on arbitrary thresholds and numbers.
- [x] Allow non-floats in`extrapolate_fdm`.
- [x] Automatically copy `README.md` to `docs/src/index.md` in `docs/make.jl`.

Additions:
- [x] More elaborate accuracy tests.
- [x] Expansion of README. 
- [x] Resolves #129: Add a `max_range` keyword that specifies how far `f` is allowed to be evaluated from `x`:
```julia
julia> m = central_fdm(5, 1, max_range=1e-6); grad(m, sum, ones(5))
```

Breaking changes:
- [x] Remove `max_step` keyword argument.


oxinabox:
Maybe StaticArrays.jl, rather than tuples?

wesselb:
With the current spaghetti code, allocations are down to zero, which achieves a 4x speed-up compared to `master`:

```julia
julia> using FiniteDifferences, BenchmarkTools

julia> m = central_fdm(10, 2, adapt=2)
@benchmark m(sinFiniteDifferenceMethod:
  order of method:       10
  order of derivative:   2
  grid:                  [-5, -4, -3, -2, -1, 1, 2, 3, 4, 5]
  coefficients:          [-0.011298500881834215, 0.11119929453262786, -0.4830357142857143, 1.1558201058201059, -0.7726851851851851, -0.7726851851851851, 1.1558201058201059, -0.4830357142857143, 0.11119929453262786, -0.011298500881834215]

julia> m(sin, 1) + sin(1)
3.8624659026709196e-13

julia> @benchmark $m(sin, 1)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     916.167 ns (0.00% GC)
  median time:      926.521 ns (0.00% GC)
  mean time:        1.027 μs (0.00% GC)
  maximum time:     3.898 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     48
```

`master`:
```julia
julia> m(sin, 1) + sin(1)
2.97983859809392e-13

julia> @benchmark $m(sin, 1)
BenchmarkTools.Trial:
  memory estimate:  1.80 KiB
  allocs estimate:  92
  --------------
  minimum time:     3.824 μs (0.00% GC)
  median time:      3.942 μs (0.00% GC)
  mean time:        4.196 μs (1.39% GC)
  maximum time:     207.103 μs (97.42% GC)
  --------------
  samples:          10000
  evals/sample:     8
```

test/methods.jl
```diff
@@ -1,70 +1,46 @@
...
+        @test FiniteDifferenceMethod([-2, 0, 5], 1)(sin, 1) ≈ cos(1)
+    end
 
-        @test m(10, 2)(foo.f, T(1)) isa T
-        if T == Float64
-            @test m(10, 2)(foo.f, T(1)) ≈ T(foo.d2)
-        end
+    @testset "Test allocations" begin
+        m = central_fdm(5, 2, adapt=2)
+        @test (@benchmark $m(sin, 1)).allocs == 0
```
willtebbutt, line 43: To make this run more quickly, you could instead write something like:
```julia
@test (@benchmark $m(sin, 1); samples=1, evals=1).allocs == 0
```
This severly limits the number of function evaluations that `BenchmarkTools` is allowed to perform but doesn't seem to effect the allocation estimate too much.

test/methods.jl
```diff
@@ -1,70 +1,46 @@
...
+        # tolerance.
+        types = [Float32, Float64]
+
+        # The different functions to evaluate (`.f`), their first derivative at 1 (`.d1`),
+        # and second derivative at 1 (`.d2`).
+        foos = [
+            (f=sin, d1=cos(1), d2=-sin(1)),
+            (f=exp, d1=exp(1), d2=exp(1)),
+            (f=abs2, d1=2, d2=2),
+            (f=x -> sqrt(max(x + 1,0)), d1=0.5 / sqrt(2), d2=-0.25 / 2^(3/2)),
```
willtebbutt, line 16: Would you consider adding the function suggested in #124 here, or does it continue to be tested elsewhere?

test/methods.jl
```diff
@@ -1,70 +1,46 @@
...
+        ]
+
+        # Test all combinations of the above settings, i.e. differentiate all functions using
+        # all methods and data types.
+        @testset "foo=$(foo.f), method=$m, type=$(T)" for foo in foos, m in methods, T in types
+            @testset "method-order=$order" for order in [1, 2, 3, 4, 5]
+                @test m(order, 0, adapt=2)(foo.f, T(1)) isa T
+                @test m(order, 0, adapt=2)(foo.f, T(1)) == T(foo.f(1))
```
willtebbutt, line 24:
```suggestion
                @test m(order, 0; adapt=2)(foo.f, T(1)) isa T
                @test m(order, 0; adapt=2)(foo.f, T(1)) == T(foo.f(1))
```

test/methods.jl
```diff
@@ -104,27 +79,27 @@ import FiniteDifferences: estimate_magitude, estimate_roundoff_error
...
-        @test FiniteDifferences._is_symmetric([2, 1, -1, -2], negate_half=true)
-        @test FiniteDifferences._is_symmetric([2, 1, 0, -1, -2], negate_half=true)
-        @test FiniteDifferences._is_symmetric([2, 1, 4, -1, -2], negate_half=true)
+        @test !FiniteDifferences._is_symmetric(SVector{4}(2, 1, -1, -2))
+        @test FiniteDifferences._is_symmetric(SVector{4}(2, 1, -1, -2), negate_half=true)
+        @test FiniteDifferences._is_symmetric(SVector{5}(2, 1, 0, -1, -2), negate_half=true)
+        @test FiniteDifferences._is_symmetric(SVector{5}(2, 1, 4, -1, -2), negate_half=true)
         @test !FiniteDifferences._is_symmetric(
-            [2, 1, 4, -1, -2],
+            SVector{5}(2, 1, 4, -1, -2),
```
willtebbutt, line 102:
```suggestion
            SVector{5}(2, 1, 4, -1, -2);
```

test/methods.jl
```diff
@@ -154,12 +129,12 @@ import FiniteDifferences: estimate_magitude, estimate_roundoff_error
         end
     end
 
-    @testset "Derivative of cosc at 0 (#124)" begin
-        @test central_fdm(5, 1)(cosc, 0) ≈ -(pi ^ 2) / 3 atol=1e-9
-        @test central_fdm(10, 1, adapt=3)(cosc, 0) ≈ -(pi ^ 2) / 3 atol=5e-14
+    @testset "Derivative of cosc at 0 (issue #124)" begin
```
willtebbutt, line 132: Ah, I see that this is tested here. Is there a reason not to included it in your standard test suite?

test/methods.jl
```diff
@@ -154,12 +129,12 @@ import FiniteDifferences: estimate_magitude, estimate_roundoff_error
...
+        @test central_fdm(5, 1)(cosc, 0) ≈ -(pi ^ 2) / 3 atol=5e-9
+        @test central_fdm(10, 1, adapt=3)(cosc, 0) ≈ -(pi ^ 2) / 3 atol=5e-13
     end
 
-    @testset "Derivative of a constant (#125)" begin
+    @testset "Derivative of a constant (issue #125)" begin
```
willtebbutt, line 137: Same here

src/methods.jl
```diff
@@ -1,102 +1,147 @@
...
-    coefs::C
+struct AdaptedFiniteDifferenceMethod{
+    P,
+    Q,
+    E<:FiniteDifferenceMethod
+} <: FiniteDifferenceMethod{P,Q}
+    grid::SVector{P,Int}
+    coefs::SVector{P,Float64}
+    coefs_neighbourhood::NTuple{3,SVector{P,Float64}}
+    condition::Float64
```
willtebbutt, line 87: What's the reason for making these `Float64`s rather than making this more generic?

src/methods.jl
```diff
@@ -1,102 +1,147 @@
...
+    perform adaptation for this finite difference method.
 """
-struct FiniteDifferenceMethod{G<:AbstractVector, C<:AbstractVector, E<:Function}
-    grid::G
-    q::Int
-    coefs::C
+struct AdaptedFiniteDifferenceMethod{
+    P,
+    Q,
+    E<:FiniteDifferenceMethod
```
willtebbutt, line 82:
```suggestion
    P, Q, E<:FiniteDifferenceMethod
```

src/methods.jl
```diff
@@ -1,102 +1,147 @@
...
-    grid::AbstractVector,
+    grid::Vector{Int},
     q::Int;
-    condition::Real=DEFAULT_CONDITION
+    condition::Real=DEFAULT_CONDITION,
+    factor::Real=DEFAULT_FACTOR,
+    max_range::Real=Inf
 )
     p = length(grid)
+    grid = SVector{p}(grid)  # Internally, we work with static vectors.
```
willtebbutt, line 128: Would it be better to provide two methods of `FiniteDifferenceMethod`, one of which accepts any old vector and converts it to an `SVector`? I can imagine a performance-conscious user might want to provide an `SVector` to start with.

src/methods.jl
```diff
@@ -1,102 +1,147 @@
...
+- `factor::Real`: Factor number. See [`DEFAULT_FACTOR`](@ref).
+- `max_range::Real=Inf`: Maximum distance that a function is evaluated from the input at
+    which the derivative is estimated.
 
 # Returns
 - `FiniteDifferenceMethod`: Specified finite difference method.
 """
 function FiniteDifferenceMethod(
-    grid::AbstractVector,
+    grid::Vector{Int},
```
willtebbutt, line 121:
```suggestion
    grid::Vector{Int};
```

src/methods.jl
```diff
@@ -1,102 +1,147 @@
...
     q::Int;
-    condition::Real=DEFAULT_CONDITION
+    condition::Real=DEFAULT_CONDITION,
+    factor::Real=DEFAULT_FACTOR,
+    max_range::Real=Inf
```
willtebbutt, line 125:
```suggestion
    max_range::Real=Inf,
```

src/methods.jl
```diff
@@ -175,26 +209,42 @@ FiniteDifferenceMethod:
...
 
 julia> fdm(sin, 1, 1e-3) - cos(1)  # Check the error.
 -1.7741363933510002e-13
 ```
 """
-@inline function (m::FiniteDifferenceMethod)(f::Function, x::Real, h::Real)
+function (m::FiniteDifferenceMethod{P,Q})(
+    f::TF,
+    x::Real,
+    step::Real
```
willtebbutt, line 221:
```suggestion
    f::TF, x::Real, step::Real
```

src/methods.jl
```diff
@@ -175,26 +209,42 @@ FiniteDifferenceMethod:
...
-    h::Real
-) where T<:AbstractFloat
-    return sum(
-        i -> convert(T, m.coefs[i]) * f(T(x + h * m.grid[i])),
-        eachindex(m.grid)
-    ) / h^m.q
+    step::Real
+) where {TF<:Function,T<:AbstractFloat}
+    xs = x .+ T(step) .* m.grid
+    return f.(xs)
```
willtebbutt, line 236:
```suggestion
    return f.(x .+ T(step) .* m.grid)
```
?

src/methods.jl
```diff
@@ -175,26 +209,42 @@ FiniteDifferenceMethod:
...
+
+function _eval_method(
+    m::FiniteDifferenceMethod{P,Q},
+    fs::SVector{P,TF},
+    x::T,
+    step::Real,
+    coefs::SVector{P,Float64}
+) where {P,Q,TF,T<:AbstractFloat}
+    coefs = T.(coefs)
+    return sum(fs .* coefs) ./ T(step)^Q
```
willtebbutt, line 247:
```suggestion
    return sum(fs .* T.(coefs)) ./ T(step)^Q
```
?

src/methods.jl
wesselb: Hmm, because the coefficients in `_coefs_mults` are also converted to `Float64`s. I guess we could add a type parameter to parametrise the operating data type for internal computations. Would you be in favour of that?

src/methods.jl
wesselb: Yes, definitely.

src/methods.jl
wesselb: For some reason this substitution causes the method to allocate. I'm not 100% sure why. I've added a comment to explain the current formatting of the code.

src/methods.jl
willtebbutt: I can't say that I'm overly fussed either way. Provided that you don't think it's a problem, we could just go with`Float64` until need to generalise.

src/methods.jl
willtebbutt: Hmm yeah, odd. This comment definitely suffices. Thanks!

test/methods.jl
wesselb: Hmm, I like to think of this test set as one that tests for correctness rather than expected accuracy. We can similarly aggregate a testset that tests for accuracy. I think that would be good.

test/methods.jl
wesselb: Good suggestion, thanks!

src/methods.jl
wesselb: Yes, definitely. Added!

willtebbutt (approved):
Went through this with Wessel over Zoom -- seems reasonable. Would like another review before merging preferably though.

Project.toml
```diff
@@ -1,23 +1,24 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.11.7"
+version = "0.12.0"
```
oxinabox, line 3: Why is this breaking?

README.md
```diff
@@ -21,16 +21,15 @@ At some point in the future they might merge, or one might depend on the other.
 Right now here are the differences:
 
  - FiniteDifferences.jl supports basically any type, where as FiniteDiff.jl supports only array-ish types
- - FiniteDifferences.jl supports higher order approximation
+ - FiniteDifferences.jl supports higher order approximation and step size adaptation
  - FiniteDiff.jl is carefully optimized to minimize allocations
```
oxinabox, line 25: We can delete that now?
Or change it to say "both are..."

Do we have a "both are" section?
Maybe we should.
"Both are carefully optimized to minimize allocations, both are expertly designed to give high accuracy finite differencing based derivative estimates."

README.md
```diff
@@ -39,17 +38,149 @@ julia> central_fdm(5, 1)(sin, 1) - cos(1)
...
 ```julia
 julia> central_fdm(5, 2)(sin, 1) + sin(1)
--4.367495254342657e-11
+-8.767431225464861e-11
+```
+
+To obtain better accuracy, you can increase the order of the method:
+
+```julia
+julia> central_fdm(12, 2)(sin, 1) + sin(1)
```
oxinabox, line 70: Can we write this as `-(-sin(x))`?

README.md
```diff
@@ -39,17 +38,149 @@ julia> central_fdm(5, 1)(sin, 1) - cos(1)
...
+```
+
+Similarly, you can estimate higher order derivatives:
+
+```julia
+julia> extrapolate_fdm(central_fdm(5, 4), sin, 1)[1] - sin(1)
+-1.626274487942503e-5
+```
+
+In this case, the accuracy can be improved by lowering (making closer to `1`)
```
oxinabox, line 152: You mean closer to ,0?

docs/src/index.md
```diff
@@ -1,51 +1,273 @@
 # FiniteDifferences.jl: Finite Difference Methods
```
oxinabox, line 1: I don't care for the copy paste.
One bit will get outdated.

Can we do the copy step during the build of the docs?
In `docs/make.jl`

Though it's not changed by this PR so maybe van just open an issue?
And make sure this PR stays right

src/FiniteDifferences.jl
```diff
@@ -5,6 +5,7 @@ module FiniteDifferences
     using Printf
     using Random
     using Richardson
+    using StaticArrays
```
oxinabox, line 8: Weird indent?

test/methods.jl
```diff
@@ -1,70 +1,80 @@
...
+        @test FiniteDifferenceMethod([-2, 0, 5], 1)(sin, 1) ≈ cos(1)
+    end
 
-        @test m(10, 2)(foo.f, T(1)) isa T
-        if T == Float64
-            @test m(10, 2)(foo.f, T(1)) ≈ T(foo.d2)
-        end
+    @testset "Test allocations" begin
+        m = central_fdm(5, 2, adapt=2)
+        @test @benchmark($m(sin, 1); samples=1, evals=1).allocs == 0
```
oxinabox, line 77: Use `@ballocated` here

oxinabox (commented):
I have not yet looked at the actual code in methods.jl

README.md
wesselb: No, I mean closer to `1`. I meant lowered as in making less aggressive. I'll change the wording.

docs/src/index.md
wesselb: I've added the copy step in `docs/make.jl`. That's indeed more convenient.

Project.toml
wesselb: I figured that a minor version increase was warranted because the internals now work very differently, whic affects the behaviour of the estimates (hopefully in a good way). One breaking change is that the keyword argument `max_step` is removed.

README.md
wesselb: Hmm, perhaps FiniteDiff.jl still has the advantage on the allocation front by supporting in-place computation and this package the advantage on the robustness/accuracy front by supporting high-order methods and step size adaptation. Maybe change the sentence to "FiniteDiff.jl supports in-place computation"?

oxinabox:
we need to drop Julia-Nightly from the CI (out of scope for this PR)

docs/make.jl
```diff
@@ -1,5 +1,14 @@
 using Documenter, FiniteDifferences
 
+# Copy the README.
+repo_root = dirname(dirname(@__FILE__))
+cp(
+    joinpath(repo_root, "README.md"),
+    joinpath(repo_root, "docs", "src", "index.md");
+    # `index.md` will already exist, so we set `force=true` to overwrite it.
```
oxinabox, line 8: we should delete and add it to `.gitignore` it, or someone will re-add it by mistake,.

src/methods.jl
oxinabox: I have a MWE
https://github.com/JuliaLang/julia/issues/39151

src/methods.jl
```diff
@@ -175,68 +214,108 @@ FiniteDifferenceMethod:
...
-    p > 20 && throw(DomainError(p, "order of the method (`p`) is too large to be computed"))
-    return
 end
 
-const _COEFFS_CACHE = Dict{Tuple{AbstractVector{<:Real}, Integer}, Vector{Float64}}()
+function _coefs(grid, p, q)
+    # For high precision on the `\`, we use `Rational`, and to prevent overflows we use
+    # `BigInt`. At the end we go to `Float64` for fast floating point math, rather than
+    # rational math.
+    C = [Rational{BigInt}(g^i) for i in 0:(p - 1), g in grid]
```
oxinabox, line 268: was `Int128` not doing it for us?

src/methods.jl
```diff
@@ -175,68 +214,108 @@ FiniteDifferenceMethod:
...
+    x = zeros(Rational{BigInt}, p)
+    x[q + 1] = factorial(big(q))
+    return SVector{p}(Float64.(C \ x))
+end
+
+const _COEFFS_MULTS_CACHE = Dict{
+    Tuple{SVector,Integer},
```
oxinabox, line 275:
```suggestion
    Tuple{SVector,Integer},  #keys: (grid, q)
```

src/methods.jl
```diff
@@ -175,68 +214,108 @@ FiniteDifferenceMethod:
...
+    Tuple{SVector,Tuple{Vararg{SVector}},Float64,Float64}
```
oxinabox, line 276:
```suggestion
    # values: (coefs, coefs_neighbourhood, ∇f_magnitude_mult, f_error_mult)
    Tuple{SVector,Tuple{Vararg{SVector}},Float64,Float64}  
```

src/methods.jl
```diff
@@ -258,63 +335,134 @@ estimate of the derivative.
...
 
-    return h, accuracy
+function _limit_step(
+    m::FiniteDifferenceMethod, x::T, step::Real, acc::Real,
+) where {T<:AbstractFloat}
+    # First, limit the step size based on the maximum range.
+    step_max = m.max_range / maximum(abs.(m.grid))
+    if step > step_max
+        step = step_max
+        acc = NaN
```
oxinabox, line 401: is it documented somewhere what this means?

Also should we just stop computing accuracy?
Its never used nor exposed to the user AFAICT

oxinabox (approved):

src/methods.jl
wesselb: No, `Int128` unfortunately hit overflow for orders bigger than 12.

src/methods.jl
wesselb: The function `estimate_step` is exposed to the user, which promises to also estimate the accuracy. You're right that `NaN` isn't documented, which I'll add.

src/methods.jl
```diff
@@ -1,102 +1,146 @@
...
+    P, Q, E<:FiniteDifferenceMethod
+} <: FiniteDifferenceMethod{P,Q}
+    grid::SVector{P,Int}
+    coefs::SVector{P,Float64}
+    coefs_neighbourhood::NTuple{3,SVector{P,Float64}}
+    condition::Float64
+    factor::Float64
+    max_range::Float64
+    ∇f_magnitude_mult::Float64
+    f_error_mult::Float64
```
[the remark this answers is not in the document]
wesselb: Certainly another good alternative. The downside is that this would introduce a bunch of accessor methods like

```julia
_get_factor(m::UnadaptedFiniteDifferenceMethod) = m.factor
_get_factor(m::AdaptedFiniteDifferenceMethod) = _get_factor(m.inner)
```
which are not necessary right now. I also considered this, but decided to copy, because that seemed like the simpler option and the duplication appears fairly harmless.

Happy to go with whichever option you find better.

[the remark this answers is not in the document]
oxinabox: Fair enough let's merge as is.
Can always change later

wesselb:
@oxinabox The ChainRules integration test fails because ChainRules and ChainRulesUtils both have a `[compat]` entry that does not allow version 0.12.0 for FiniteDifferences. I tried to fix this by removing the entry from the Project.toml in the cloned ChainRules repo in the GitHub workflow, but that fails because also ChainRulesUtils restricts FiniteDifferences. I there a way to tell Julia to ignore requirements for FiniteDifferences?

oxinabox:
There is not.
We don't need them to pass.

wesselb:
Fair enough, I guess we can just check this PR locally. Running the ChainRules tests locally, it appears that some tests fail due to singularities, but those failures are resolved by changing

```julia
const _fdm = central(5, 1)
```
to

```julia
const _fdm = central(5, 1; max_range=1e-2)
```

in `src/ChainRulesTestUtils.jl`. I'll make a PR.

Are we happy to merge this?

oxinabox:
> Are we happy to merge this?

Yes.

Issue #133, opened 2021-01-05
Bug when using Matrix as differential for Adjoint


sethaxen:
#131 introduced a bug when differentials of `Adjoint` are regular matrices. Minimal example:

```julia
julia> using FiniteDifferences, LinearAlgebra

julia> n, T = 3, ComplexF64;

julia> x = randn(T, n);

julia> y = adjoint(x);

julia> ȳ_adjoint = adjoint(randn(T, n))
1×3 adjoint(::Vector{ComplexF64}) with eltype ComplexF64:
 -0.0351965+0.649028im  0.868707+0.237612im  -1.45535-0.149933im

julia> ȳ_mat = collect(ȳ_adjoint);

julia> ȳ_composite = FiniteDifferences.Composite{typeof(y)}(parent=parent(ȳ_adjoint));

julia> ȳ_adjoint' # expected
3-element Vector{ComplexF64}:
 -0.03519650654436998 - 0.6490275763507989im
   0.8687071179992576 - 0.23761165442658064im
  -1.4553505347699434 + 0.14993263689809244im

julia> j′vp(central_fdm(5, 1), adjoint, ȳ_adjoint, x)[1] # fine
3-element Vector{ComplexF64}:
 -0.035196506544367144 - 0.6490275763508034im
    0.8687071179992422 - 0.2376116544265735im
   -1.4553505347699498 + 0.14993263689809289im

julia> j′vp(central_fdm(5, 1), adjoint, ȳ_composite, x)[1] # fine
3-element Vector{ComplexF64}:
 -0.035196506544367144 - 0.6490275763508034im
    0.8687071179992422 - 0.2376116544265735im
   -1.4553505347699498 + 0.14993263689809289im

julia> j′vp(central_fdm(5, 1), adjoint, ȳ_mat, x)[1] # conjugated
3-element Vector{ComplexF64}:
 -0.03519650654436785 + 0.6490275763508094im
    0.868707117999241 + 0.23761165442657906im
  -1.4553505347699507 - 0.1499326368980859im
```

I've stared at the PR but can't figure out where it went wrong. Or maybe the issue is that a regular matrix is not an acceptable cotangent for an `Adjoint`. This was caught by the ChainRules test suite.
@willtebbutt I also noticed that PR disabled testing of `to_vec` for many complex types. Was that intentional?

willtebbutt:
[...]

PR #134, opened 2021-01-05
WIP: Adjoint problems

willtebbutt:
As @sethaxen pointed out in #133, #131 accidentally disabled a number of tests, and introduced a weird bug. This PR currently resolves the first of these problems, and I'm presently working on the second.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.11.6"
+version = "0.11.7"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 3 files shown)


willtebbutt:
@sethaxen could you confirm that this resolves the problem?

sethaxen:
> @sethaxen could you confirm that this resolves the problem? I'm quite nervous about this because I don't really understand what's going on 😬 .

Well this is bizarre. Before it was correct for `Adjoint` and `Composite` differentials but conjugated for `Matrix`. Now this is correct for `Matrix` and `Adjoint` differentials but is conjugated for `Composite`

willtebbutt:
Booooooooooooooooo. I wonder whether it ever consistent prior to #131 ?

sethaxen:
Oooooh, not it wasn't. `Composite` has always been wrong for `Adjoint`. ChainRules worked around that by checking that the pullback produced the same output for `Matrix` and `Composite` cotangents but only passed the `Matrix` to `rrule_test`.
I think then this looks good to go.

sethaxen (approved):

Issue #132, opened 2021-01-03
Some ambiguity in `to_vec`


willtebbutt:
Presently we have implementations of `to_vec` like [this](https://github.com/JuliaDiff/FiniteDifferences.jl/blob/cfef6f8ac5800a4d4cb62f63013c2becc6b6ab7e/src/to_vec.jl#L97) instead of implementations that look like [this](https://github.com/JuliaDiff/FiniteDifferences.jl/blob/cfef6f8ac5800a4d4cb62f63013c2becc6b6ab7e/src/to_vec.jl#L105).

IIRC the reason for this is that if you define things in the latter fashion then you'll find if you do something like
```julia
A = randn(5, 4)
A_vec, _ = to_vec(A)
At_vec, _ = to_vec(transpose(A))
```
then `A_vec == At_vec`. My suspicion is that at some point in time, we were trying to compare the equality of two things based on their `to_vec` representation and it came up that things weren't being compared properly. We do a similar thing with `Diagonal` etc.

Unfortunately, it seems quite straightforward to construct examples which don't admit such a straightforward resolution. For a trivial example, consider that
```julia
A = randn(5)
A == (A, ) # is false -- a `Tuple` is never equal to a `Vector`
```
However, since `to_vec(::Tuple)` simply concatenates `to_vec` of each of the fields, we'll find that
```julia
to_vec(A)[1] == to_vec((A, ))[1]
```
This leads me to believe that it's likely not possible to define `to_vec` such that equality is preserved.

As such, I propose that we
1. embrace, and explicitly document, that knowing whether `to_vec(x)[1] == to_vec(y)[1]` in general tells you nothing about whether `x == y`, and vice versa. We should point this out to explicitly warn people away from trying to check whether two things are equal based on their `to_vec` representation.
2. refactor some of the things (like `Tranpose` and `Adjoint`) to use the struct representation, which is more computationally efficient.
3. `to_vec` should purely be thought of as a way to make `FiniteDifferences`' life easier. We might even consider ceasing to export it.

Now that @oxinabox has added some new approximate equality checking to `ChainRulesTestUtils`, this ought to be less of a problem anyway.

willtebbutt:
I also suspect that this change would make it much safer to rely on recursive definitions of `to_vec`. For example, we ought to be able to define a single (maybe `@generated`) version of `to_vec` for all `struct`-types. Presently we're unable to do this because we're trying to preserve equality.

[...]

PR #136, opened 2021-01-10
Add ChainRules integration test

sethaxen:
As proposed in https://github.com/JuliaDiff/FiniteDifferences.jl/issues/132#issuecomment-757393572, this PR adds an integration test workflow for ChainRules (copied from the equivalent workflow used by ChainRulesCore). ChainRules extensively uses FiniteDifferences, and sometimes regressions here are not caught by our test suite but do cause ChainRules failures. This should help us catch these.

### DIFF
--- a/.github/workflows/IntegrationTestChainRules.yml
+++ b/.github/workflows/IntegrationTestChainRules.yml
@@ -1,0 +1,30 @@
+name: IntegrationTestChainRules 
+on:
+  push:
+    branches: [master]
+    tags: [v*]
+  pull_request:
+
+jobs:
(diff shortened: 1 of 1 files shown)


willtebbutt (approved):
LGTM -- I guess this is copied directly from `ChainRulesCore`?

Would be good if @oxinabox could give the thumbs up before we merge though.

sethaxen:
> LGTM -- I guess this is copied directly from `ChainRulesCore`?

Yes that's right.

oxinabox (approved):

PR #139, opened 2021-01-15
More reverse dependency tests

oxinabox:
These will probably a bit flaky still with versions since its quiet possibly  to make a semver incompatible change here that is also not supported by ChainRulesTestUtils
which will mean these still give boy who cried wolf failures during tests. as it will not be able to find a version that meets the 3 requirements of Exactly this version of FiniteDifferences, `#master` of the downstream package, and some permitted version of ChainRulesTestUtils

Still worth doing


mattBrzezinski (approved):

wesselb (approved):
Nice, this LGTM.

PR #140, opened 2021-01-15
Fallback `to_vec` intended for structs

mzgubic:
Currently the user has to define a `to_vec` function for every new struct. This PR adds a fallback for structs.

Thanks @willtebbutt for the implementation, I've just added a line to check against structs with no fieldnames (`String`s and singleton types for example). I couldn't think of anything else messing this up, but I'll have another think about it on Monday.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.12.0"
+version = "0.12.1"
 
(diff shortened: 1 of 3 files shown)


wesselb (approved):
This LGTM.

mzgubic:
We don't mind the Nightly Julia tests failing right? It looks like there are some inference issues

wesselb:
I don't think we do, but I'm not sure. IIRC the Julia nightlies have been failing for some time now on this error.

PR #142, opened 2021-01-20
Fix CI Cache bug. add x86 jobs

fchorney:
There was a CI Cache bug affecting the previous ci files (found in a different project).

there was also an exclude for macOS x86 jobs, but there were no x86 jobs actually running. Added x86 back in assuming that's what we want here.

### DIFF
--- a/.github/workflows/JuliaNightly.yml
+++ b/.github/workflows/JuliaNightly.yml
@@ -16,15 +16,16 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - 'nightly'
+          - "nightly"
         os:
           - ubuntu-latest
           - macOS-latest
           - windows-latest
         arch:
           - x64
-        # 32-bit Julia binaries are not available on macOS
+          - x86
         exclude:
+          # 32-bit Julia binaries are not available on macOS
           - os: macOS-latest
             arch: x86
     steps:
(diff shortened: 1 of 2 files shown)


fchorney:
Do we want nightly running on pull requests here?

fchorney:
Do we also want an explicit 1.5 job on here for when `1` ends up going to 1.6?

mattBrzezinski:
> Do we also want an explicit 1.5 job on here for when `1` ends up going to 1.6?

It does exist in our internal packages, probably worth having it explicit as our stable version.

.github/workflows/JuliaNightly.yml
```diff
@@ -16,32 +16,34 @@ jobs:
...
           - os: macOS-latest
             arch: x86
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
         with:
           version: ${{ matrix.version }}
-          arch: ${{ matrix.arch }}
-      - uses: actions/cache@v1
+          arich: ${{ matrix.arch }}
```
mattBrzezinski, line 36: I don't think this was meant to be changed.

```suggestion
          arch: ${{ matrix.arch }}
```

oxinabox:
> Do we want nightly running on pull requests here?

Out of scope for this PR.
Can make another later if we want to change it

PR #145, opened 2021-02-08
Infinite recursion in wrapped arrays

### LINKED ISSUE #141: infinite recursion of to_vec for wrapped arrays
I didn't look at the implementation details of `to_vec` but I wanted to report a bug we encountered in
https://github.com/FluxML/NNlib.jl/pull/272 when applying `to_vec` to a custom array wrapper. 
```
142
AutoDiff: Error During Test at /home/runner/work/NNlib.jl/NNlib.jl/test/batchedmul.jl:235
143
  Got exception outside of a @test
144
  StackOverflowError:
145
  Stacktrace:
146
       [1] to_vec(x::BatchedAdjoint{Float64, Array{Float64, 3}})
147
         @ FiniteDifferences ~/.julia/packages/FiniteDifferences/7NROH/src/to_vec.jl:56
148
       [2] to_vec(x::Base.ReshapedArray{Float64, 1, BatchedAdjoint{Float64, Array{Float64, 3}}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}, Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}})
149
         @ FiniteDifferences ~/.julia/packages/FiniteDifferences/7NROH/src/to_vec.jl:69
150
       [3] to_vec(x::BatchedAdjoint{Float64, Array{Float64, 3}})
151
         @ FiniteDifferences ~/.julia/packages/FiniteDifferences/7NROH/src/to_vec.jl:57--- the last 2 lines are repeated 39990 more times ---
```
We worked around the issue defining:
```julia
FiniteDifferences.to_vec(x::BatchedAdjoint) = FiniteDifferences.to_vec(collect(x))
```
Could this be used as a generic fallback?

mzgubic:
Closes #141

### DIFF
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -54,7 +54,7 @@ function to_vec(x::AbstractVector)
 end
 
 function to_vec(x::AbstractArray)
-    x_vec, from_vec = to_vec(vec(x))
+    x_vec, from_vec = to_vec(collect(vec(x)))
 
     function Array_from_vec(x_vec)
         return oftype(x, reshape(from_vec(x_vec), size(x)))
--- a/test/to_vec.jl
+++ b/test/to_vec.jl
@@ -18,6 +18,9 @@ struct FillVector <: AbstractVector{Float64}
     len::Int
 end
 
+Base.size(x::FillVector) = (x.len,)
+Base.getindex(x::FillVector, n::Int) = x.x
+
 # For testing Composite{ThreeFields}
 struct ThreeFields
     a
@@ -32,8 +35,14 @@ struct Nested
     y::Singleton
 end
 
-Base.size(x::FillVector) = (x.len,)
-Base.getindex(x::FillVector, n::Int) = x.x
+# for testing wrapped arrays
+struct WrappedArray{T, V<:AbstractMatrix{T}} <: AbstractMatrix{T}
+    a::V
+end
+
+Base.size(w::WrappedArray) = size(w.a)
+Base.getindex(w::WrappedArray, i, j) = getindex(w.a, i, j)
+Base.convert(::Type{WrappedArray{T, Array{T, N}}}, a::Array{T, N}) where {T, N} = WrappedArray(a)
 
 function test_to_vec(x::T; check_inferred = true) where {T}
     check_inferred && @inferred to_vec(x)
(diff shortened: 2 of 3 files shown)


test/to_vec.jl
```diff
@@ -18,6 +18,9 @@ struct FillVector <: AbstractVector{Float64}
     len::Int
 end
 
+Base.size(x::FillVector) = (x.len,)
```
mzgubic, line 21: This has just been moved

mzgubic:
Do we care about Julia nightly failed tests?

willtebbutt:
> Do we care about Julia nightly failed tests?

I don't think so. @nickrobinson251 any thoughts?

I wonder whether this is a good way to solve this problem? In particular, it requires that whichever `AbstractArray` is `to_vec`ed define `convert` from `Array` to whatever type it is.

Tackling the `BlockDiagonal` example directly, do we have enough information in the type to know how to `convert` (required in `from_vec`) a `Matrix{<:Real}` to a `BlockDiagonal`? Is it not under-specified, since we don't know the block sizes in convert?

src/to_vec.jl
```diff
@@ -54,7 +54,7 @@ function to_vec(x::AbstractVector)
 end
 
 function to_vec(x::AbstractArray)
-    x_vec, from_vec = to_vec(vec(x))
+    x_vec, from_vec = to_vec(collect(vec(x)))
```
nickrobinson251, line 57:
```suggestion
    # `collect` to avoid recursion in wrapped arrays
    x_vec, from_vec = to_vec(collect(vec(x)))
```

nickrobinson251 (approved):
LGTM

Tests on Nightly version of Julia don't matter (they look entirely unrelated to this PR)

mzgubic:
Having slept on it it's probably best to not merge this and just define a `to_vec` for array types directly for now. I think the current error is easier to debug than what we would get from this.

The long term plan is to get rid of `to_vec` anyway in favour of ChainRules differential types anyway right?

willtebbutt:
> The long term plan is to get rid of to_vec anyway in favour of ChainRules differential types anyway right?

It is indeed the plan.

gdalle:
Can I suggest revisiting this PR? I found it was needed to make DifferentiationInterface.jl work with FiniteDifferences.jl, see https://github.com/JuliaDiff/FiniteDifferences.jl/issues/141#issuecomment-2015046426

PR #146, opened 2021-02-16
Only use the diagonal values in Diag, not the whole Matrix

mzgubic:


mzgubic:
Oh, this breaks ChainRules tests

oxinabox:
@willtebbutt

willtebbutt:
Ah, yes, this relates to [this discussion](https://github.com/JuliaDiff/FiniteDifferences.jl/issues/132) -- I'm in favour of doing this, but we probably need to resolve some of the stuff in that conversation.

mzgubic:
https://github.com/JuliaDiff/FiniteDifferences.jl/issues/186

PR #148, opened 2021-03-31
Fix 1.6 CI

mzgubic:
It looks like there are some type inference issues with `reduce` in 1.6
https://github.com/JuliaLang/julia/issues

So removing these checks for now to make CI work again


oxinabox:
https://github.com/JuliaLang/julia/issues/40277

oxinabox (approved):

PR #155, opened 2021-04-08
copy `rand_tangent(::BigFloat)` from `ChainRulesTestUtils`

mzgubic:
starting https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/51

by moving the only difference https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/132

### DIFF
--- a/src/rand_tangent.jl
+++ b/src/rand_tangent.jl
@@ -13,7 +13,8 @@ rand_tangent(rng::AbstractRNG, x::Integer) = DoesNotExist()
 
 rand_tangent(rng::AbstractRNG, x::T) where {T<:Number} = randn(rng, T)
 
-# ref: https://github.com/JuliaLang/julia/issues/17629
+# TODO: right now Julia don't allow `randn(rng, BigFloat)` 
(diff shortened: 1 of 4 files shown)


mzgubic:
@oxinabox  this is the first step towards a single `rand_tangent` function

src/rand_tangent.jl
```diff
@@ -13,6 +13,9 @@ rand_tangent(rng::AbstractRNG, x::Integer) = DoesNotExist()
 
 rand_tangent(rng::AbstractRNG, x::T) where {T<:Number} = randn(rng, T)
 
+# ref: https://github.com/JuliaLang/julia/issues/17629
```
oxinabox, line 16:
```suggestion
# TODO: right now Julia don't allow `randn(rng, BigFloat)` 
# see: https://github.com/JuliaLang/julia/issues/17629
```

oxinabox (approved):
I am not sure that moving things into FiniteDifferences is the right way, but it certainly is less suffering than having it in both.

Issue #149, opened 2021-04-06
Should `rand_tangent(2:3)` return `DoesNotExist()`?


mzgubic:
I can't think of any use-case where the current
```julia
julia> ChainRulesTestUtils.rand_tangent(2:3)
Composite{UnitRange{Int64}}(start = ChainRulesCore.DoesNotExist(), stop = ChainRulesCore.DoesNotExist())
```
would be the better option, can anyone else?

There are potentially other indices that could be updated, such as `CartesianIndex` and `(:)`.

willtebbutt:
Coming back to this: I feel strongly that patching this by making this a `DoesNotExist` is a bad idea, and that we ought to make `to_vec` work properly for subtypes of `AbstractArray`.

mzgubic:
Is it possible to write `to_vec` for a `UnitRange`/`StepRange`/`StepRangeLen`? I mean, it's easy to turn it into a vector, but once you perturb the vector it is no longer possible to construct a range from it, right?

In other words these various ranges seem to be outside of the domain of the `to_vec` function. Perhaps the fallback `to_vec` methods for `AbstractArray/Vector` should instead be `StridedArray/Vector`?

willtebbutt:
> Is it possible to write to_vec for a UnitRange/StepRange/StepRangeLen? I mean, it's easy to turn it into a vector, but once you perturb the vector it is no longer possible to construct a range from it, right?

So this is another example of the "is it an array, or is it a struct?" question. In this case I think it's clear cut that they ought to be treated as structs. If you take that view, there's really no problem -- they're just structs and we utilise the recursive definition of `to_vec` like we would for any other struct.

> In other words these various ranges seem to be outside of the domain of the to_vec function. Perhaps the fallback to_vec methods for AbstractArray/Vector should instead be StridedArray/Vector?

I think that we should try to implement these tighter typing constraints anyway because, as we can see in this example, `to_vec(::AbstractArray)` doesn't work.

oxinabox:
`rand_tangent(2:3)`  *should* return `DoesNotExist()`.
particularly since `rand_tangent(1)` returns that.
We treat integers as non-pertubable.
thus `AbstractArray{<:Interger}` should also be non-perturable.

willtebbutt:
Agreed -- either that or a `Composite` with `DoesNotExist()`s in.

mzgubic:
[...]

PR #156, opened 2021-04-20
Constrain type in `to_vec(::AbstractArray/Vector)`

mzgubic:
I've looked into whether we can simply constrain the `to_vec` generic `AbstractVector/Array` to `Strided`, and it almost works.

It's just these `ChainRules` tests which need the `SubArray` in the `Union` (I didn't investigate why the `SubArray`s inside the `Strided` `Union` do not contain this case):
https://github.com/JuliaDiff/ChainRules.jl/blob/38caf4bfdb8af616fcbe7626d10699608af21904/test/rulesets/Base/arraymath.jl#L36-L42 

@willtebbutt this seems to solve a lot of issues that we were seeing (https://github.com/JuliaDiff/FiniteDifferences.jl/issues/149, https://github.com/JuliaDiff/FiniteDifferences.jl/issues/141).

Is there a reason we haven't changed this before? I guess partly because it is breaking, but are there any bad things that can happen if we go ahead with this? It seems like using the struct fallback is a better choice for most things that subtype `AbstractArray`s?

### DIFF
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -40,7 +40,7 @@ function to_vec(x::T) where {T}
     return v, structtype_from_vec
 end
 
-function to_vec(x::Union{SubArray, Base.ReshapedArray, StridedVector})
+function to_vec(x::DenseVector)
     x_vecs_and_backs = map(to_vec, x)
     x_vecs, backs = first.(x_vecs_and_backs), last.(x_vecs_and_backs)
     function Vector_from_vec(x_vec)
@@ -53,7 +53,7 @@ function to_vec(x::Union{SubArray, Base.ReshapedArray, StridedVector})
     return x_vec, Vector_from_vec
 end
 
-function to_vec(x::Union{SubArray, Base.ReshapedArray, StridedArray})
+function to_vec(x::DenseArray)
     x_vec, from_vec = to_vec(vec(x))
 
     function Array_from_vec(x_vec)
@@ -75,10 +75,11 @@ function to_vec(x::Base.ReshapedArray{<:Any, 1})
     return x_vec, ReshapedArray_from_vec
 end
 
-# To return a SubArray we would endup needing to copy the `parent` of `x` in `from_vec`
-# which doesn't seem particularly useful. So we just convert the view into a copy.
-# we might be able to do something more performant but this seems good for now.
-to_vec(x::Base.SubArray) = to_vec(copy(x))
+function to_vec(x::SubArray)
+    x_vec, from_vec = to_vec(x.parent)
+    SubArray_from_vec(x_vec) = view(from_vec(x_vec), x.indices...)
+    return x_vec, SubArray_from_vec
+end
 
 function to_vec(x::T) where {T<:LinearAlgebra.AbstractTriangular}
     x_vec, back = to_vec(Matrix(x))
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -40,7 +40,7 @@ function to_vec(x::T) where {T}
     return v, structtype_from_vec
 end
 
-function to_vec(x::StridedVector)
+function to_vec(x::Union{SubArray, Base.ReshapedArray, StridedVector})
     x_vecs_and_backs = map(to_vec, x)
     x_vecs, backs = first.(x_vecs_and_backs), last.(x_vecs_and_backs)
     function Vector_from_vec(x_vec)
@@ -53,7 +53,7 @@ function to_vec(x::StridedVector)
     return x_vec, Vector_from_vec
 end
 
-function to_vec(x::Union{SubArray, StridedArray})
+function to_vec(x::Union{SubArray, Base.ReshapedArray, StridedArray})
     x_vec, from_vec = to_vec(vec(x))
 
     function Array_from_vec(x_vec)
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -40,7 +40,7 @@ function to_vec(x::T) where {T}
     return v, structtype_from_vec
 end
 
-function to_vec(x::AbstractVector)
+function to_vec(x::StridedVector)
     x_vecs_and_backs = map(to_vec, x)
     x_vecs, backs = first.(x_vecs_and_backs), last.(x_vecs_and_backs)
     function Vector_from_vec(x_vec)
@@ -53,7 +53,7 @@ function to_vec(x::AbstractVector)
     return x_vec, Vector_from_vec
 end
 
-function to_vec(x::AbstractArray)
+function to_vec(x::Union{SubArray, StridedArray})
     x_vec, from_vec = to_vec(vec(x))
 
     function Array_from_vec(x_vec)
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -34,8 +34,8 @@ function to_vec(x::T) where {T}
     v, vals_from_vec = to_vec(vals)
     function structtype_from_vec(v::Vector{<:Real})
         val_vecs = vals_from_vec(v)
-        vals = map((b, v) -> b(v), backs, val_vecs)
-        return T(vals...)
+        values = map((b, v) -> b(v), backs, val_vecs)
+        return T(values...)
     end
     return v, structtype_from_vec
 end
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -63,8 +63,9 @@ function to_vec(x::DenseArray)
     return x_vec, Array_from_vec
 end
 
-
 # Some specific subtypes of AbstractArray.
+to_vec(x::SubArray) = to_vec(copy(x))
+
 function to_vec(x::Base.ReshapedArray{<:Any, 1})
     x_vec, from_vec = to_vec(parent(x))
     function ReshapedArray_from_vec(x_vec)
@@ -73,12 +74,6 @@ function to_vec(x::Base.ReshapedArray{<:Any, 1})
     end
 
     return x_vec, ReshapedArray_from_vec
-end
-
-function to_vec(x::SubArray)
-    x_vec, from_vec = to_vec(x.parent)
-    SubArray_from_vec(x_vec) = view(from_vec(x_vec), x.indices...)
-    return x_vec, SubArray_from_vec
 end
 
 function to_vec(x::T) where {T<:LinearAlgebra.AbstractTriangular}
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -64,8 +64,6 @@ function to_vec(x::DenseArray)
 end
 
 # Some specific subtypes of AbstractArray.
-to_vec(x::SubArray) = to_vec(copy(x))
-
 function to_vec(x::Base.ReshapedArray{<:Any, 1})
     x_vec, from_vec = to_vec(parent(x))
     function ReshapedArray_from_vec(x_vec)
@@ -75,6 +73,11 @@ function to_vec(x::Base.ReshapedArray{<:Any, 1})
 
     return x_vec, ReshapedArray_from_vec
 end
+
+# To return a SubArray we would endup needing to copy the `parent` of `x` in `from_vec`
+# which doesn't seem particularly useful. So we just convert the view into a copy.
+# we might be able to do something more performant but this seems good for now.
+to_vec(x::Base.SubArray) = to_vec(copy(x))
 
 function to_vec(x::T) where {T<:LinearAlgebra.AbstractTriangular}
     x_vec, back = to_vec(Matrix(x))
--- a/test/to_vec.jl
+++ b/test/to_vec.jl
@@ -34,6 +34,16 @@ struct Nested
     y::Singleton
 end
 
+# For testing generic subtypes of AbstractArray
+struct WrapperArray{T, N, A<:AbstractArray{T, N}} <: AbstractArray{T, N}
+    data::A
+end
+function WrapperArray(a::AbstractArray{T, N}) where {T, N}
+    return WrapperArray{T, N, AbstractArray{T, N}}(a)
+end
+Base.size(a::WrapperArray) = size(a.data)
+Base.getindex(a::WrapperArray, inds...) = getindex(a.data, inds...)
+
 function test_to_vec(x::T; check_inferred=true) where {T}
     check_inferred && @inferred to_vec(x)
     x_vec, back = to_vec(x)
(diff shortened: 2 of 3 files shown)


willtebbutt:
> Is there a reason we haven't changed this before? I guess partly because it is breaking, but are there any bad things that can happen if we go ahead with this? It seems like using the struct fallback is a better choice for most things that subtype AbstractArrays?

I think the answer is just that no one has taken the time to fix it -- I had personally imagined that it would be annoying to have to modify a bunch of existing code in ChainRules, but if that's not the case, then great.

Regarding bad things happening -- there might be a couple of surprised users, and perhaps the occassional issue asking why an `AbstractArray` tangent isn't being generated, but that's about it.

src/to_vec.jl
```diff
@@ -53,7 +53,7 @@ function to_vec(x::StridedVector)
     return x_vec, Vector_from_vec
 end
 
-function to_vec(x::Union{SubArray, StridedArray})
+function to_vec(x::Union{SubArray, Base.ReshapedArray, StridedArray})
```
willtebbutt, line 56: Maybe the parent array of `SubArray` should be restricted to e.g. `StridedArray` or something?

mzgubic: We want e.g. a
```julia
julia> sa = @view randn(T, 10)[1:4]
julia> typeof(sa)
SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}
```
to dispatch on the current `SubArray` implementation.  Unfortunately `StridedArray` is a `Union` of several more specific `SubArrays` (not just in the parent argument)

So we have two options:
1) Narrow the dispatch of the current `SubArray` (very verbose*)
2) Loosen the dispatch for `StridedArray` to `Union{StridedArray, SubArray}` so that the current `SubArray` becomes more specific.

And similarly for the `ReshapedArray`.

I don't particularly like either option but it's probably better than the status quo.

*Number 1) looks like this, I want to nope out of that 😂 

```julia
StridedSubArray = Union{
SubArray{T, N, A, I, true} where {T, N, A<:DenseArray, I<:Union{Tuple{Vararg{Real, N} where N}, Tuple{AbstractUnitRange, Vararg{Any, N} where N}}}, DenseArray}, MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}, N} where N}},
SubArray{T, 1, A, I, L} where {A<:Union{Base.ReinterpretArray{T, N, S, A, IsReshaped} where {T, N, A<:Union{SubArray{T, N, A, I, true} where {T, N, A<:DenseArray, I<:Union{Tuple{Vararg{Real, N} where N}, Tuple{AbstractUnitRange, Vararg{Any, N} where N}}}, DenseArray}, IsReshaped, S},
SubArray{T, N, A, I, true} where {T, N, A<:DenseArray, I<:Union{Tuple{Vararg{Real, N} where N}, Tuple{AbstractUnitRange, Vararg{Any, N} where N}}}, DenseArray}, MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}, N} where N}}, DenseArray}, I<:Tuple{Vararg{Union{Int64, AbstractRange{Int64}, Base.AbstractCartesianIndex, Base.ReshapedArray{T, N, A, Tuple{}} where {T, N, A<:AbstractUnitRange}}, N} where N}, L}} where T)
}
```

mzgubic: @willtebbutt how do you feel about that?

willtebbutt: The thing I'm concerned about is if someone attempts to `to_vec` something like
```julia
view(Diagonal(randn(5)), 1:5, 1:3)
```

> *Number 1) looks like this, I want to nope out of that 😂

Lol yeah, that's really not fun.

Is there a reason that we can't go with something like
```julia
Union{StridedArray, SubArray{T, D, <:StridedArray}}
```
and just recurse into `ReshapedArray`s?

mzgubic: The dispatch for the `Diagonal` example is the same (the dedicated `SubArray` method) whether we constrain the parent of the `SubArray` to be strided or not.

The example breaks, but because `SparseArray.SparseMatrixCSC` does not have a `to_vec` defined.

willtebbutt: Wait, we are able to `to_vec` a sparse array? That makes no sense to me.

mzgubic: No no, sorry, I must have confused you. This is what happens with the `Diagonal` example:
```julia
julia> vd = view(Diagonal(randn(5)), 1:5, 1:3)
5×3 view(::Diagonal{Float64, Vector{Float64}}, 1:5, 1:3) with eltype Float64:
 0.168618   0.0        0.0
 0.0       -0.796772   0.0
 0.0        0.0       -1.16312
 0.0        0.0        0.0
 0.0        0.0        0.0

julia> v, back = to_vec(vd)
([5.0, 3.0, 1.0, 2.0, 3.0, 4.0, 1.0, 2.0, 3.0, 0.16861792127500824, -0.796772203387132, -1.1631233934390923], FiniteDifferences.var"#structtype_from_vec#29"{SparseMatrixCSC{Float64, Int64}, FiniteDifferences.var"#Tuple_from_vec#45"{NTuple{5, Int64}, NTuple{5, Int64}, NTuple{5, typeof(identity)}}, Tuple{FiniteDifferences.var"#Real_from_vec#22", FiniteDifferences.var"#Real_from_vec#22", typeof(identity), typeof(identity), typeof(identity)}}(FiniteDifferences.var"#Tuple_from_vec#45"{NTuple{5, Int64}, NTuple{5, Int64}, NTuple{5, typeof(identity)}}((1, 2, 6, 9, 12), (1, 1, 4, 3, 3), (identity, identity, identity, identity, identity)), (FiniteDifferences.var"#Real_from_vec#22"(), FiniteDifferences.var"#Real_from_vec#22"(), identity, identity, identity)))

julia> back(v)
ERROR: MethodError: no method matching SparseMatrixCSC{Float64, Int64}(::Float64, ::Float64, ::Vector{Float64}, ::Vector{Float64}, ::Vector{Float64})
Closest candidates are:
  SparseMatrixCSC{Tv, Ti}(::Integer, ::Integer, ::Vector{Ti}, ::Vector{Ti}, ::Vector{Tv}) where {Tv, Ti<:Integer} at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/SparseArrays/src/sparsematrix.jl:26
Stacktrace:
 [1] (::FiniteDifferences.var"#structtype_from_vec#29"{SparseMatrixCSC{Float64, Int64}, FiniteDifferences.var"#Tuple_from_vec#45"{NTuple{5, Int64}, NTuple{5, Int64}, NTuple{5, typeof(identity)}}, Tuple{FiniteDifferences.var"#Real_from_vec#22", FiniteDifferences.var"#Real_from_vec#22", typeof(identity), typeof(identity), typeof(identity)}})(v::Vector{Float64})
   @ FiniteDifferences ~/JuliaEnvs/ChainRules.jl/dev/FiniteDifferences/src/to_vec.jl:38
 [2] top-level scope
   @ REPL[85]:1
```

The dispatch (and the error message) is the same in both cases, whereas we have
```julia
function to_vec(x::Union{SubArray{T, D, <:StridedArray}, Base.ReshapedArray, StridedArray}) where {T, D}
```
or
```julia
function to_vec(x::Union{SubArray, Base.ReshapedArray, StridedArray})
```

willtebbutt: Lol -- I wonder how we wound up with that sparse array...

willtebbutt: > ([5.0, 3.0, 1.0, 2.0, 3.0, 4.0, 1.0, 2.0, 3.0, 0.16861792127500824, -0.796772203387132, -1.1631233934390923], FiniteDifferences.var"#structtype_from_vec#29"{SparseMatrixCSC{Float64, Int64}, FiniteDifferences.var"#Tuple_from_vec#45"{NTuple{5, Int64}, NTuple{5, Int64}, NTuple{5, typeof(identity)}}, Tuple{FiniteDifferences.var"#Real_from_vec#22", FiniteDifferences.var"#Real_from_vec#22", typeof(identity), typeof(identity), typeof(identity)}}(FiniteDifferences.var"#Tuple_from_vec#45"{NTuple{5, Int64}, NTuple{5, Int64}, NTuple{5, typeof(identity)}}((1, 2, 6, 9, 12), (1, 1, 4, 3, 3), (identity, identity, identity, identity, identity)), (FiniteDifferences.var"#Real_from_vec#22"(), FiniteDifferences.var"#Real_from_vec#22"(), identity, identity, identity)))

And where does the `[5.0, 3.0, 1.0...]` come from 🤔

willtebbutt: Oh, it's the method on line 81
```julia
to_vec(x::Base.SubArray) = to_vec(copy(x))
```

willtebbutt: Yeah, if you remove that method and the `SubArray`s from the unions, it seems to work, which makes sense to me.

mzgubic: So what's the upshot? Keep as is or solve the `SparseMatrixCSC` issue as well? I feel like that is a separate PR?

willtebbutt: I'm confused. My understanding of the situation is
1. the SparseMatrixCSC stuff was an artifact of something going wrong
2. removing `to_vec` for `Base.SubArray`, and removing `SubArray` from the Union makes everything work.
Is this not the case?

mzgubic: The status is:
- All tests pass with the current PR
- Removing the `to_vec(Base.SubArray)` breaks many tests
- `test_to_vec(view(Diagonal(randn(5)), 1:5, 1:3); check_inferred=false)`, (the thing you suggested should be checked) breaks on master (`StackOverflowError`), on the current PR (`SparseMatrixCSC` error), and on the current PR with `SubArray` removed from the Union (`SparseMatrixCSC` error).

willtebbutt: Ahh I see.

I've run it further locally, and found that the following seems to resolve everything for me:

1. restrict the array methods to `DenseVector` and `DenseArray`, from `Union{StridedArray...}`
2. add a specific method for `SubArray`:
```julia
function to_vec(x::SubArray)
    x_vec, from_vec = to_vec(x.parent)
    SubArray_from_vec(x_vec) = view(from_vec(x_vec), x.indices...)
    return x_vec, SubArray_from_vec
end
```
if you don't have this, it thinks that the integer fields should also be `to_vec`ed.
3. comment out the `SubArray`-specific method

mzgubic: Oh yeah that's much better. Are we happy with that?

willtebbutt: I'm happy with it, but I feel that we should get an additional review. Maybe @oxinabox or @sethaxen ?

mzgubic: Actually with this one ChainRules test fails, and the reason is that the current implementation returns the whole parent vector
```julia
julia> a = rand(7);

julia> b = @view a[2:5];

julia> v, back = to_vec(b);

julia> v
7-element Vector{Float64}:
 0.9586867155061674
 0.8025760289977295
 0.9784238101274141
 0.16440182965461236
 0.17784328126321847
 0.01704930118685577
 0.8408143603848348
```

Keeping the old version below works just fine. No issues now `StridedArray` has been replaced by `DenseArray` (where we used to have `AbstractArray`)
```julia
to_vec(x::SubArray) = to_vec(copy(x))
```

willtebbutt:
This is looking pretty good now I think. Do we need to bump the minor version, or would a patch bump suffice?

mzgubic:
I've made sure ChainRules tests are fine but technically it is still a breaking change, no?

willtebbutt:
> I've made sure ChainRules tests are fine but technically it is still a breaking change, no?

Im not sure which bit is really breaking. Could you elaborate?

mzgubic:
```julia
julia> subtypes(AbstractArray)
31-element Vector{Any}:
 AbstractRange
 Base.LogicalIndex
 Base.ReinterpretArray
 Base.ReshapedArray
 Base.SCartesianIndices2
 BitArray
 CartesianIndices
 Core.Compiler.AbstractRange
 Core.Compiler.BitArray
 Core.Compiler.LinearIndices
 DenseArray
 LinearAlgebra.AbstractQ
 LinearAlgebra.AbstractTriangular
 LinearAlgebra.Adjoint
 LinearAlgebra.Bidiagonal
 LinearAlgebra.Diagonal
 LinearAlgebra.Hermitian
 LinearAlgebra.LQPackedQ
 LinearAlgebra.SymTridiagonal
 LinearAlgebra.Symmetric
 LinearAlgebra.Transpose
 LinearAlgebra.Tridiagonal
 LinearAlgebra.UpperHessenberg
 LinearIndices
 PermutedDimsArray
 SparseArrays.AbstractSparseArray
 StaticArrays.StaticArray
 StaticArrays.TrivialView
 SubArray
 SuiteSparse.CHOLMOD.FactorComponent
 Test.GenericArray
```

I guess unless a special `to_vec` is defined for these types (as well as any subtypes of `AbstractArray` in other packages, e.g. `BlockDiagonal`) they will now change to the fallback method. I suppose this is breaking though it could also be interpreted as a bugfix since they mostly broke before anyway?

willtebbutt:
> I suppose this is breaking though it could also be interpreted as a bugfix since they mostly broke before anyway?

This is my view -- if anyone opens an issue indicating that we broke their code, I'm more than happy to add a `to_vec` for the array type that broke.

A third opinion would be great though.

test/to_vec.jl
```diff
@@ -67,8 +66,8 @@ end
         test_to_vec(UpperTriangular(randn(T, 13, 13)))
         test_to_vec(Diagonal(randn(T, 7)))
         test_to_vec(DummyType(randn(T, 2, 9)))
-        test_to_vec(SVector{2, T}(1.0, 2.0))
-        test_to_vec(SMatrix{2, 2, T}(1.0, 2.0, 3.0, 4.0))
+        test_to_vec(SVector{2, T}(1.0, 2.0); check_inferred = false)
+        test_to_vec(SMatrix{2, 2, T}(1.0, 2.0, 3.0, 4.0); check_inferred = false)
```
oxinabox, line 70:
```suggestion
        test_to_vec(SVector{2, T}(1.0, 2.0); check_inferred=false)
        test_to_vec(SMatrix{2, 2, T}(1.0, 2.0, 3.0, 4.0); check_inferred=false)
```
Bluetyle spacing in kwargs

oxinabox:
I think we need a test for a custom wrapper array, something like is like  NamedDimsArray.
but without the names.
Probably
```
struct WrapperArray{T<:AbstractArray} <: AbstractArray
    data::T
end
# minimal methods that just delegate
``` 

and make sure that still acts in a way we are happy with.

test/to_vec.jl
mzgubic: changed in other places as well now

oxinabox (approved):
I think we can say this is a patch

PR #157, opened 2021-04-28
unbroken type inference

mzgubic:
`reduce` has [broken type inference on 1.6.0](https://github.com/JuliaLang/julia/issues/40277). This has been fixed in 1.6.1.


oxinabox (approved):
Is it worth making sure to not test on 1.6.0?
or leaving a comment to that effect?

I guess not, since we should assume people are using the latest patch release of the programming language if running tests?

Issue #153, opened 2021-04-08
`rand_tangent(::Colon)` should return `DoesNotExist()`


mzgubic:
first mentioned in https://github.com/JuliaDiff/ChainRules.jl/pull/396#discussion_r609619038

willtebbutt:
Is this true? It's a struct, and it doesn't have any fields, so `NO_FIELDS` seems appropriate, this is what [this rule](https://github.com/JuliaDiff/FiniteDifferences.jl/blob/c31952549df3607c4809d141c533d83d0015e23a/src/rand_tangent.jl#L26) does.

However, `NO_FIELDS` is currently equal to `Zero`.

It seems we have a contradiction. @oxinabox ?

willtebbutt:
> I do agree that DoesNotExist makes sense though.

In light of reviewing https://github.com/JuliaDiff/ChainRulesCore.jl/issues/175, I retract this. An empty `Composite` makes more sense to me.

PR #158, opened 2021-04-29
make rand_tangent of struct with no perturbable fields return DoesNotExist

oxinabox:
Closes https://github.com/JuliaDiff/FiniteDifferences.jl/issues/149
This is my preferred alternative to  https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/142#issuecomment-829287623

Note that it is **not** the case that if all my fields do not exist then i do not exist,
but it is the case that if all my fields are not perturbable then I am not pertablable.
Which is what `DoesNotExist` actually means, and we should rename it.

### DIFF
--- a/src/rand_tangent.jl
+++ b/src/rand_tangent.jl
@@ -33,7 +33,12 @@ function rand_tangent(rng::AbstractRNG, x::T) where {T}
         tangents = map(field_names) do field_name
             rand_tangent(rng, getfield(x, field_name))
         end
-        return Composite{T}(; NamedTuple{field_names}(tangents)...)
+        if all(tangent isa DoesNotExist for tangent in tangents)
+            # if none of my fields can be perturbed then I can't be perturbed
+            return DoesNotExist()
+        else
+            Composite{T}(; NamedTuple{field_names}(tangents)...)
+        end
     else
         return NO_FIELDS
     end
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.12.3"
+version = "0.12.4"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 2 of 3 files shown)


src/rand_tangent.jl
```diff
@@ -33,7 +33,12 @@ function rand_tangent(rng::AbstractRNG, x::T) where {T}
...
         end
-        return Composite{T}(; NamedTuple{field_names}(tangents)...)
+        if all(tangent isa DoesNotExist for tangent in tangents)
+            # if none of my fields can be perturbed then I can't be perturbed
+            return DoesNotExist()
+        else
+            Composite{T}(; NamedTuple{field_names}(tangents)...)
+        end
     else
         return NO_FIELDS
```
oxinabox, line 43: The outer `if` here that gives us the `NO_FIELDS` would be able to be removed if we replaced `NO_FIELDS` with `DoesNotExist()`.
Since then we would be in the case that all the fields (of which there are none) subtype `DoesNotExist`.
Vacuous truth.

https://github.com/JuliaDiff/ChainRulesCore.jl/issues/175#issuecomment-829632604

mzgubic:
Also closes https://github.com/JuliaDiff/FiniteDifferences.jl/issues/153

mzgubic (approved):

mzgubic:
Actually, it won't close any of the issues, because we still have two copies of rand_tangent (one here and one in ChainRulesTestUtils).

oxinabox:
> Also closes #153

It won't close that

```julia
julia> rand_tangent(Random.GLOBAL_RNG, Colon())
ChainRulesCore.Zero()
```

It would close that if we also did
https://github.com/JuliaDiff/ChainRulesCore.jl/issues/175#issuecomment-829632604
which i think we should (which would also simplify the code in this)

> Actually, it won't close any of the issues, because we still have two copies of rand_tangent (one here and one in ChainRulesTestUtils).

suffering

mzgubic:
Oh yeah that's right, my bad. I am happy to move the code over and make it work. We discussed on slack a while back.

oxinabox:
Thanks.
Please do so.
Either location is better than both

mzgubic:
I think we can merge this now since https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/147 is in

mzgubic:
Well, actually, needs a version bump, not sure why it does not show up as conflict.

PR #162, opened 2021-05-17
rename differentials

mzgubic:
in line with: https://github.com/JuliaDiff/ChainRulesCore.jl/pull/353

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.jl.*.cov
 *.jl.mem
 /Manifest.toml
+dev/
 
 # Docs:
 docs/build/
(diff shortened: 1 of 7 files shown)


docs/Project.toml
```diff
@@ -4,3 +4,4 @@ FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 
 [compat]
 Documenter = "0.25"
+FiniteDifferences = "0.12.7"
```
oxinabox, line 7: This line shouldn't be here

oxinabox (approved):
Just one thing to fix, then merge when happy

docs/Project.toml
oxinabox:
```suggestion
```

PR #163, opened 2021-05-27
Replace `NO_FIELDS` by `NoTangent()`

mzgubic:
Follow up to JuliaDiff/ChainRulesCore.jl#358


oxinabox (approved):

oxinabox:
Closes #153

mzgubic:
Actually I think this is breaking because it changes the `rand_tangent` for a struct with no fields

oxinabox:
> Actually I think this is breaking because it changes the rand_tangent for a struct with no fields

I think that is a bug-fix. It is the thing that makes `Colon()` have the correct behavour

Hmm on that basis the change to NO_FIELDS might actually be a bug fix, and not a breaking change

src/rand_tangent.jl
```diff
@@ -44,6 +44,6 @@ function rand_tangent(rng::AbstractRNG, x::T) where {T}
             Tangent{T}(; NamedTuple{field_names}(tangents)...)
         end
     else
-        return NO_FIELDS
+        return NoTangent()
     end
```
oxinabox, line 48: We should not do this surface level change.
We actually get to remove the whole outer loop here.
we no longer need `if length(field_names) > 0`

mzgubic:
> > Actually I think this is breaking because it changes the rand_tangent for a struct with no fields
> 
> I think that is a bug-fix. It is the thing that makes `Colon()` have the correct behavour
> 
> Hmm on that basis the change to NO_FIELDS might actually be a bug fix, and not a breaking change

I am coming around to think that this PR can be passed as a bug fix, but not the ChainRulesCore one (since it affects too many downstream packages that were relying on this "buggy" behaviour)

Issue #151, opened 2021-03-24
rand_tangent(adjoint(array)) returns Composite even though Adjoint is a natural differential


mzgubic:
Presently
```julia
julia> A = rand(3, 3);

julia> adj = adjoint(A);

julia> ChainRulesTestUtils.rand_tangent(adj)
Composite{Adjoint{Float64,Array{Float64,2}}}(parent = [-0.5319309079526605 -1.2397774126992458 1.450107932767501; 0.2443697244819896 -0.5283784206385908 -0.36484818100635125; -0.18829408436246423 -1.0652797084400936 -1.0457119567770545],)
```

whereas we would like to get an Adjoint out, which is a natural differential type.

Probably just need to define a new dispatch for `rand_tangent`?

willtebbutt:
>  This package is (primarily) for testing rules humans write.

This might be true right now, but it's not where we're headed. Quite near the top of the TODO list is making `test_rrule` and `test_frule` accept an AD as an argument, so that we can use them to test arbitrary bits of code.

>  the reason we call it natural is it is the one humans use when writing rules.

I agree that humans use this package to test rules, and humans seem to like making the tangents of arrays other arrays, but they write rules that are to be _used_ inside AD.

AD much prefers `Composite`s to natural differentials -- if a human doesn't test their rule using the thing that AD likes, then they've not tested their rule properly.

Moreover, part of the point of `rand_tangent` is to make it so that humans don't have to think hard about exactly what type the AD would like, they just make have `rand_tangent` generate it for them, and then they figure out how to deal with it in their rule.

oxinabox:
I stand by the statement that we should do this.


> AD much prefers Composites to natural differentials -- if a human doesn't test their rule using the thing that AD likes, then they've not tested their rule properly.

Possibly we should have an extra flag to also test the Composite.
but for array rules, its often super useful to accept array deriviatives.
That means you can use your own weird linear operations defined in your package on them.
People want to write rules that way, we should let them.
The code is clearer when written using natural differentials.
We should sort it out so they always can write in the form they want.

And its perfectly common for them to only get array inputs, because the source of the cotangent is another rule.

Not having this means i have to write a custom tangent for like 20 tests.

willtebbutt:
Fair enough. My takeaway from this is just that we
1. need to get projections sorted, and
2. that we need to use that functionality to make it straightforward to generate tangents in pretty arbitrary forms.

PR #165, opened 2021-05-27
Make rand_tangent on adjoint an transpose return natural

oxinabox:
closes https://github.com/JuliaDiff/FiniteDifferences.jl/issues/151

### DIFF
--- a/src/rand_tangent.jl
+++ b/src/rand_tangent.jl
@@ -19,7 +19,7 @@ rand_tangent(rng::AbstractRNG, ::BigFloat) = big(randn(rng))
 
 rand_tangent(rng::AbstractRNG, x::StridedArray) = rand_tangent.(Ref(rng), x)
 rand_tangent(rng::AbstractRNG, x::Adjoint) = adjoint(rand_tangent(rng, parent(x)))
-rand_tangent(rng::AbstractRNG, x::Tangent) = tangent(rand_tangent(rng, parent(x)))
+rand_tangent(rng::AbstractRNG, x::Transpose) = transpose(rand_tangent(rng, parent(x)))
 
 function rand_tangent(rng::AbstractRNG, x::T) where {T<:Tuple}
     return Tangent{T}(rand_tangent.(Ref(rng), x)...)
--- a/src/rand_tangent.jl
+++ b/src/rand_tangent.jl
@@ -13,11 +13,13 @@ rand_tangent(rng::AbstractRNG, x::Integer) = NoTangent()
 
 rand_tangent(rng::AbstractRNG, x::T) where {T<:Number} = randn(rng, T)
 
-# TODO: right now Julia don't allow `randn(rng, BigFloat)` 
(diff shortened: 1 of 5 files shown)


src/rand_tangent.jl
```diff
@@ -13,11 +13,13 @@ rand_tangent(rng::AbstractRNG, x::Integer) = NoTangent()
...
 rand_tangent(rng::AbstractRNG, x::T) where {T<:Number} = randn(rng, T)
 
-# TODO: right now Julia don't allow `randn(rng, BigFloat)` 
+# TODO: right now Julia don't allow `randn(rng, BigFloat)`
 # see: https://github.com/JuliaLang/julia/issues/17629
 rand_tangent(rng::AbstractRNG, ::BigFloat) = big(randn(rng))
 
 rand_tangent(rng::AbstractRNG, x::StridedArray) = rand_tangent.(Ref(rng), x)
+rand_tangent(rng::AbstractRNG, x::Adjoint) = adjoint(rand_tangent(rng, parent(x)))
+rand_tangent(rng::AbstractRNG, x::Tangent) = tangent(rand_tangent(rng, parent(x)))
```
nickrobinson251, line 23: Typo?

src/rand_tangent.jl
```diff
@@ -13,11 +13,13 @@ rand_tangent(rng::AbstractRNG, x::Integer) = NoTangent()
...
```
oxinabox, line 22:
```suggestion
rand_tangent(rng::AbstractRNG, x::Transpose) = transpose(rand_tangent(rng, parent(x)))
```

test/rand_tangent.jl
```diff
@@ -24,6 +24,11 @@ using FiniteDifferences: rand_tangent
         (randn(Complex{Float32}, 5, 4), Matrix{Complex{Float32}}),
         ([randn(5, 4), 4.0], Vector{Any}),
 
+        # Wrapper Arrays
+        (randn(5, 4)', Adjoint{Float64, Matrix{Float64}},
+        (transpose(randn(5, 4)), Transpose{Float64, Matrix{Float64}},
```
oxinabox, line 29:
```suggestion
        (randn(5, 4)', Adjoint{Float64, Matrix{Float64}}),
        (transpose(randn(5, 4)), Transpose{Float64, Matrix{Float64}}),
```

oxinabox:
Added two more changes.

Firstly to stop testing exacty what type `+` returned.
Because `Adjoint  + Adjoint = Matrix` so it isn't trivially consistent, but per my last comments in #151 I don't really care that much, we have to handle that case because we have to handle that case.
Better to be pragmatic.
It's also out of the power of this package to fix.
That will be fixed with projection/rerepresentation which will be defined else where.

Secondly, this really relaxed a lot about how difference works.
I think the new behavour is intuitive but it feels more dubious.
But I want to delete `difference` anyway, so i don't think it matters (#166)

mzgubic:
LGTM after `difference` is deleted

mzgubic (approved):

PR #167, opened 2021-05-28
delete difference

### LINKED ISSUE #166: Delete `difference`?
The `difference` function is a first step towards #97.
However the rest of the steps for #97 haven't been done yet, and are not planned in the short-term.

Maintaining `difference` is kinda annoying,
and I suspect once we have the notion of projecting to different representations will want some redesigning.

It's not exported and it isn't used, so i think we can delete it as a nonbreaking change.

We can bring it back later when we need it.

oxinabox:
closes #166


willtebbutt (approved):
Seems reasonable to me. Just needs a patch bump.

PR #168, opened 2021-05-28
make rand_tangent give only nice numbers

oxinabox:
I don't think we ever cared about rand_tangent giving actual random valuis.
This rigs it so they are ones with fairly short decimal expressions.
(vs the standard for a Float of 19 charater and 42 characters for a Complex).

It is a bit of a balencing act between restricting the sample space, and getting things that are short.

I have 2 stratergies:

1. `round(k*randn(x) sigdigits=n, base=2)` this is I guess pretty general, but it resticts the space to `z*2^-n`.
2. `rand(range)` there are certain ranges of values that floating point can exactly represent. For example all Float's in  `Base` can exactly represent `-9:0.001:9` (but not `-9:0.0001:9`).

I use stratergy 2 for `Float64` and `ComplexF64` (the types we care about the most)
and strategy 1 for everything else.
Though perhaps I should just use `randn(T)` for everything else and not worry about it.
and/Or code the stratergy 2 for Float16, Float32, and the complex versions.

Contributes towards
https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/146

### DIFF
--- a/src/rand_tangent.jl
+++ b/src/rand_tangent.jl
@@ -16,7 +16,8 @@ rand_tangent(rng::AbstractRNG, x::Integer) = NoTangent()
 # Try and make nice numbers with short decimal representations for good error messages
 # while also not biasing the sample space too much
 function rand_tangent(rng::AbstractRNG, x::T) where {T<:Number}
-    return round(8randn(rng, T), sigdigits=5, base=2)
+    # multiply by 9 to give a bigger range of values tested: no so tightly clustered around 0.
+    return round(9 * randn(rng, T), sigdigits=5, base=2)
 end
 rand_tangent(rng::AbstractRNG, x::Float64) = rand(rng, -9:0.01:9)
 function rand_tangent(rng::AbstractRNG, x::ComplexF64)
@@ -24,7 +25,9 @@ function rand_tangent(rng::AbstractRNG, x::ComplexF64)
 end
 
 #BigFloat/MPFR is finicky about short numbers, this doesn't always work as well as it should
-rand_tangent(rng::AbstractRNG, ::BigFloat) = round(big(8randn(rng)), sigdigits=5, base=2)
+
+# multiply by 9 to give a bigger range of values tested: no so tightly clustered around 0.
+rand_tangent(rng::AbstractRNG, ::BigFloat) = round(big(9 * randn(rng)), sigdigits=5, base=2)
 
 rand_tangent(rng::AbstractRNG, x::StridedArray) = rand_tangent.(Ref(rng), x)
 rand_tangent(rng::AbstractRNG, x::Adjoint) = adjoint(rand_tangent(rng, parent(x)))
--- a/src/rand_tangent.jl
+++ b/src/rand_tangent.jl
@@ -16,17 +16,15 @@ rand_tangent(rng::AbstractRNG, x::Integer) = NoTangent()
 # Try and make nice numbers with short decimal representations for good error messages
 # while also not biasing the sample space too much
 function rand_tangent(rng::AbstractRNG, x::T) where {T<:Number}
-    return round(8randn(rng, T), sigdigits=6, base=2)
+    return round(8randn(rng, T), sigdigits=5, base=2)
 end
 rand_tangent(rng::AbstractRNG, x::Float64) = rand(rng, -9:0.01:9)
 function rand_tangent(rng::AbstractRNG, x::ComplexF64)
     return ComplexF64(rand(rng, -9:0.1:9), rand(rng, -9:0.1:9))
 end
 
-
-# TODO: right now Julia don't allow `randn(rng, BigFloat)`
-# see: https://github.com/JuliaLang/julia/issues/17629
-rand_tangent(rng::AbstractRNG, ::BigFloat) = big(rand_tangent(rng, Float64))
+#BigFloat/MPFR is finicky about short numbers, this doesn't always work as well as it should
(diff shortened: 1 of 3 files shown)


oxinabox:
Example output

```julia
julia> rand_tangent(Random.default_rng(), [20,1, 1.0])
3-element Vector{Float64}:
 -1.85
  5.8
  6.37

julia> rand_tangent(Random.default_rng(), [20,1, 1.0 + im])
3-element Vector{ComplexF64}:
  1.3 + 4.7im
 -3.4 - 7.2im
  5.8 - 1.8im

julia> rand_tangent(Random.default_rng(), [20,1, 1f0])
3-element Vector{Float32}:
  3.1875
  7.25
 -1.3125
 ```
 
 vs before
 
 ```julia
 julia> rand_tangent(Random.default_rng(), [20,1, 1.0])
3-element Vector{Float64}:
  0.763615021012554
  0.18907650152439567
 -1.145201534296794

julia> rand_tangent(Random.default_rng(), [20,1, 1.0 + im])
3-element Vector{ComplexF64}:
 -1.8347206012984052 - 0.49090922332726283im
  0.3283012521981687 + 0.6142419220872114im
  0.4521831726291885 + 0.3891147996324198im

julia> rand_tangent(Random.default_rng(), [20,1, 1f0])
3-element Vector{Float32}:
  0.6569605
 -1.1273828
 -1.7002666
 ```

File: src/rand_tangent.jl [hunk 1]
```diff
@@ -11,11 +13,18 @@ rand_tangent(rng::AbstractRNG, x::AbstractString) = NoTangent()
 
 rand_tangent(rng::AbstractRNG, x::Integer) = NoTangent()
 
-rand_tangent(rng::AbstractRNG, x::T) where {T<:Number} = randn(rng, T)
+# Try and make nice numbers with short decimal representations for good error messages
+# while also not biasing the sample space too much
+function rand_tangent(rng::AbstractRNG, x::T) where {T<:Number}
+    return round(8randn(rng, T), sigdigits=5, base=2)
```
mzgubic, line 19:
```suggestion
    return round(randn(rng, T), sigdigits=5, base=2)
```
Or is there a reason why we multiply by 8? Maybe adding a comment would be helpful?

src/rand_tangent.jl
```diff
@@ -11,11 +13,18 @@ rand_tangent(rng::AbstractRNG, x::AbstractString) = NoTangent()
...
+rand_tangent(rng::AbstractRNG, x::Float64) = rand(rng, -9:0.01:9)
+function rand_tangent(rng::AbstractRNG, x::ComplexF64)
+    return ComplexF64(rand(rng, -9:0.1:9), rand(rng, -9:0.1:9))
+end
 
-# TODO: right now Julia don't allow `randn(rng, BigFloat)`
-# see: https://github.com/JuliaLang/julia/issues/17629
-rand_tangent(rng::AbstractRNG, ::BigFloat) = big(randn(rng))
+#BigFloat/MPFR is finicky about short numbers, this doesn't always work as well as it should
+rand_tangent(rng::AbstractRNG, ::BigFloat) = round(big(8randn(rng)), sigdigits=5, base=2)
```
mzgubic, line 27:
```suggestion
rand_tangent(rng::AbstractRNG, ::BigFloat) = round(big(randn(rng)), sigdigits=5, base=2)
```

mzgubic (approved):
Looking forward to this being merged. Needs a version bump

src/rand_tangent.jl
oxinabox:
```suggestion
    # multiply by 9 to give a bigger range of values tested: no so tightly clustered around 0.
    return round(9randn(rng, T), sigdigits=5, base=2)
```

src/rand_tangent.jl
oxinabox:
```suggestion

# multiply by 9 to give a bigger range of values tested: no so tightly clustered around 0.
rand_tangent(rng::AbstractRNG, ::BigFloat) = round(big(9randn(rng)), sigdigits=5, base=2)
```

src/rand_tangent.jl
nickrobinson251: please can you write it as `9 * randn` so it's easier to see?

oxinabox:
This failure on SpecialFunctions.jl is real
https://github.com/JuliaDiff/FiniteDifferences.jl/pull/168/checks?check_run_id=2718333004#step:6:232

It is only equal to `atol=1e-8` not `atol=1e-9`.
I am tempted top say we should just relax that default in ChainRulesTestUtils before we merge this?

test/rand_tangent.jl
```diff
@@ -86,4 +86,14 @@ using FiniteDifferences: rand_tangent
...
         @test x + (rand_tangent(x) + rand_tangent(x)) isa typeof(x)
     end
+
+    # Julia 1.6 changed to using Ryu printing algorithm and seems better at printing short
+    VERSION > v"1.6" && @testset "niceness of printing" begin
+        for i in 1:50
+            @test length(string(rand_tangent(1.0))) <= 6
+            @test length(string(rand_tangent(1.0 + 1.0im))) <= 12
+            @test length(string(rand_tangent(1f0))) <= 12
+            @test length(string(rand_tangent(big"1.0"))) <= 9
```
oxinabox, line 96:
```suggestion
            @test length(string(rand_tangent(big"1.0"))) <= 12
```

PR #171, opened 2021-06-04
Fix the random seed during testing

### LINKED ISSUE #169: "niceness of printing" test fails on Julia 1 MacOS and Windows
Jobs:
https://github.com/JuliaDiff/FiniteDifferences.jl/runs/2733930843?check_suite_focus=true#step:6:73
https://github.com/JuliaDiff/FiniteDifferences.jl/runs/2733930843?check_suite_focus=true#step:6:73

Seems related to https://github.com/JuliaDiff/FiniteDifferences.jl/pull/168

oxinabox:
Should close #169 
the problem is printing of BigFloats only mostly works right using out method.
I am pretty sure this is a bug in julia / MPFR

So we can set the seed to stop ourselves from getting unlucky

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -14,6 +14,8 @@ struct Foo
    c::Any
 end
 
+Random.seed!(1)
+
 @testset "FiniteDifferences" begin
     include("rand_tangent.jl")
     include("methods.jl")


mzgubic (approved):

oxinabox:
Possibly should be using `digits` rather than `sigdigits` also.
That can be a seperate PR.

Issue #170, opened 2021-06-04
recommend robust method for iteratively calculated functions


tpapp:
I have a computation that solves an implicit equation using iterative numerical methods (bisection, Brent, etc). I would like to use FiniteDifferences.jl for unit testing partials I write for ChainRulesCore.

The problem is that since the *calculated* function is not technically continuous, using eg `central_fdm(5, 1)` gives me bogus results (occasionally off by a factor of 5--10). I had some success with a high number of grid points, eg `central_fdm(20, 1)`, but too much adaptation (eg 2 steps) makes things go bad again.

I would appreciate recommendations on how to deal with this in a robust way for unit testing. Since it is for testing, it does not have to be super-fast, just robust.

wesselb:
Hey @tpapp,

How do the results look like it you set `factor` to a high number, e.g. `central_fdm(5, 1; factor=1e4)`?

tpapp:
Improved, compared to `central_fdm(5, 1)`, but `central_fdm(20, 1)` is still better (2-3 digits more accuracy).

`factor = 1e6` works fine though. Is that what I should do in this case, instead of more gridpoints?

wesselb:
If `factor=1e6` works fine, then that would be the more robust solution, I think. `factor` controls how much error derives from the function evaluations and is relative to the machine epsilon. You can even try something like `central_fdm(10, 1; factor=1e6)`.

oxinabox:
> I would like to use FiniteDifferences.jl for unit testing partials I write for ChainRulesCore.

Aside: If you are not already I recomment using ChainRulesTestUtils, which wraps FiniteDifferences.jl.

The default fdm in ChainRulesTestUtils is `central_fdm(5, 1; max_range=1e-2)`
That might not work for you (you can pass your own fdm into `test_rrule` to overwrite

wesselb:
Hey @tpapp! Would you be able to confirm whether you've found a satisfactory way to compute finite differences? If so, we'll add a recommendation to the README. :)

tpapp:
@wesselb: I am in the process of finishing up the unit tests for the relevant package, it seems to be working for now but I will get back here once I am done.

tpapp:
@wesselb: I am done, and I can confirm that `central_fdm(5, 1; factor = 1e6)` works reasonably for my purposes. I can even combine it with `max_range` to bound away from singularities.

You may close this issue, or keep it as a reminder for the readme change if you prefer.

[...]

PR #174, opened 2021-06-10
Add section about numerical noise to README

wesselb:
See title and #170

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ This package was formerly called FDM.jl. We recommend users of FDM.jl [update to
 
 * [Scalar Derivatives](#scalar-derivatives)
     - [Dealing with Singularities](#dealing-with-singularities)
+    - [Dealing with Numerical Noise](#dealing-with-numerical-noise)
     - [Richardson Extrapolation](#richardson-extrapolation)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -136,6 +137,36 @@ julia> central_fdm(5, 1, max_range=9e-1)(ε -> logdet(x .+ ε .* v), 0) - sum(1
...
+```julia
+julia> central_fdm(5, 1)(sin_noisy, 1) - cos(1)
+-0.027016678790599657
+```
+
+which is a terrible performance.
+To deal with this, you can set the keyword argument `factor`, which specifies
+numerical noise on the function evaluation relative to the machine epsilon.
+In this example, the noise lies in `[-2, 2]` (roughly) and the machine epsilon
+is `eps(Float64) ≈ 2.22e-16`, so `factor = 2e-6 / 2e-16 = 1e10` should be
```
nickrobinson251, line 159: sorry i didn't quite follow: where does the `2e-6` term come from?

wesselb: I've rewritten the sentence. Is this more clear now?

nickrobinson251: Yep, thanks!

nickrobinson251 (approved):

PR #181, opened 2021-07-16
try to handle structs with custom constructors

mzgubic:
Before we would see something like:
```julia
julia> cca = CustomConstructorArray(rand(2, 3));

julia> v, b = to_vec(cca); b(v)
T = CustomConstructorArray{Float64, 2, Matrix{Float64}}
ERROR: MethodError: no method matching CustomConstructorArray{Float64, 2, Matrix{Float64}}(::Matrix{Float64})
Stacktrace:
 [1] (::FiniteDifferences.var"#structtype_from_vec#27"{CustomConstructorArray{Float64, 2, Matrix{Float64}}, FiniteDifferences.var"#Tuple_from_vec#43"{Tuple{Int64}, Tuple{Int64}, Tuple{typeof(identity)}}, Tuple{FiniteDifferences.var"#Array_from_vec#32"{Matrix{Float64}, typeof(identity)}}})(v::Vector{Float64})
   @ FiniteDifferences ~/JuliaEnvs/test/alex/dev/FiniteDifferences/src/to_vec.jl:39
 [2] top-level scope
   @ REPL[13]:1
```
because there is a custom constructor. Not the best solution (one could imagine trying to figure out what the constructor is), but an improvement over the current state.

### DIFF
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -21,12 +21,8 @@ end
 # Base case -- if x is already a Vector{<:Real} there's no conversion necessary.
 to_vec(x::Vector{<:Real}) = (x, identity)
 
-# get around the outer constructors
-macro _force_construct(T, argsplat_ex)
-    @assert argsplat_ex.head == :(...)
-    args = first(argsplat_ex.args)
-    return esc(Expr(:splatnew, T, args))
-end
+# get around the constructors and make the type directly
+@generated _force_construct(T, args...) = Expr(:new, :T, Any[:(args[$i]) for i in 1:length(args)]...)
 
 # Fallback method for `to_vec`. Won't always do what you wanted, but should be fine a decent
 # chunk of the time.
@@ -45,7 +41,7 @@ function to_vec(x::T) where {T}
         try
             T(values...)
         catch MethodError
-            return @_force_construct(T, values...)
+            return _force_construct(T, values...)
         end
     end
     return v, structtype_from_vec
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -22,11 +22,14 @@ end
 to_vec(x::Vector{<:Real}) = (x, identity)
 
 # get around the constructors and make the type directly
-@generated function _force_construct(T, args...)
-    return if VERSION >= v"1.3"
-        Expr(:splatnew, :T, :args)
-    else
-        Expr(:new, :T, Any[:(args[$i]) for i in 1:length(args)]...)
+# Note this is moderately evil accessing julia's internals
+if VERSION >= v"1.3"
+    @generated function _force_construct(T, args...)
+        return Expr(:splatnew, :T, :args)
+    end
+else
+    @generated function _force_construct(T, args...)
+        return Expr(:new, :T, Any[:(args[$i]) for i in 1:length(args)]...)
     end
 end
 
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -23,7 +23,7 @@ to_vec(x::Vector{<:Real}) = (x, identity)
 
 # get around the constructors and make the type directly
 @generated function _force_construct(T, args...)
-    return if VERSION >= 1.3
+    return if VERSION >= v"1.3"
         Expr(:splatnew, :T, :args)
     else
         Expr(:new, :T, Any[:(args[$i]) for i in 1:length(args)]...)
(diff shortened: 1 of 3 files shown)


oxinabox:
One can bypass inner constructors.
Which might be suitable to do in this case?
is this as valid as assuming a non-parametric constructor exists?

I think it makes sense to bypass constructors because we are working with the fields directly

```julia
macro force_construct(T, args...)
          return esc(Expr(:new, T, args...))
end
```


```
julia> struct CustomConstructorArray{T, N, A<:AbstractArray{T, N}} <: AbstractArray{T, N}
           data::A
           function CustomConstructorArray(data::A) where {T, N, A<:AbstractArray{T, N}}
               return new{T, N, A}(data)
           end
       end

julia> Base.size(a::CustomConstructorArray) = size(a.data)

julia> Base.getindex(a::CustomConstructorArray, inds...) = getindex(a.data, inds...)

julia> macro force_construct(T, args...)
          return esc(Expr(:new, T, args...))
       end
@force_construct (macro with 1 method)

julia> x = CustomConstructorArray([1.0 2.0; 3.0 4.0])
2×2 CustomConstructorArray{Float64, 2, Matrix{Float64}}:
 1.0  2.0
 3.0  4.0

julia> y = @force_construct(typeof(x), [5.0 6.0; 7.0 8.0])
2×2 CustomConstructorArray{Float64, 2, Matrix{Float64}}:
 5.0  6.0
 7.0  8.0

```

mzgubic:
Wow, what is this `:new` magic? I mean I know it's role in the inner constructor - but not sure how it works outside that.

In any case, we need to splat the arguments, so the macro would have to be a bit more complicated. I'd like to give it a go if you think it's the right path forward

oxinabox:
> Wow, what is this :new magic? I mean I know its role in the inner constructor - but not sure how it works outside that.

As I understand it, it is the thing that powers the `new` insider the inner constructor.
The parser generates that `Expr(;new, current_scopes_type, args...)`  based on the current scope.
And then outside of that context it doesn't (thus preventing evil like ... us).

I think it is the right path.
It does not play nice with integer fields that need to  become floating point fields.
But nothing does.
so feels right

src/to_vec.jl
```diff
@@ -21,6 +21,13 @@ end
 # Base case -- if x is already a Vector{<:Real} there's no conversion necessary.
 to_vec(x::Vector{<:Real}) = (x, identity)
 
+# get around the outer constructors
+macro _force_construct(T, argsplat_ex)
+    @assert argsplat_ex.head == :(...)
```
oxinabox, line 26: Maybe also we assert that it has only got 1 arg?
```suggestion
    @assert Meta.isexpr(argsplat_ex, :(...), 1)
```

mzgubic: It can't have two, right? Would that not have errored earlier?

oxinabox:
It seems like 1.3 introduced splatnew

@simeonschaub  says for 1.0 we need to do

```julia
 @generated force_construct(T, args...) = Expr(:new, :T, Any[:(args[$i]) for i in 1:length(args)]...)
```
but after 1.3 we can do
```
@generated v(T, args...) = Expr(:splatnew, :T, :args)
```

src/to_vec.jl
```diff
@@ -21,6 +21,15 @@ end
 # Base case -- if x is already a Vector{<:Real} there's no conversion necessary.
 to_vec(x::Vector{<:Real}) = (x, identity)
 
+# get around the constructors and make the type directly
+@generated function _force_construct(T, args...)
+    return if VERSION >= v"1.3"
```
nickrobinson251, line 26: should the `if VERSION` be inside or outside of the function? (i've no idea if it makes a difference here, but curious if it does)

src/to_vec.jl
oxinabox: idk.
Assertions are like excutable comments,
so also including that info seems useful

src/to_vec.jl
oxinabox: Putting it on the outside will be easier on the compiler i think.

src/to_vec.jl
```diff
@@ -21,6 +21,15 @@ end
 # Base case -- if x is already a Vector{<:Real} there's no conversion necessary.
 to_vec(x::Vector{<:Real}) = (x, identity)
 
+# get around the constructors and make the type directly
+@generated function _force_construct(T, args...)
```
oxinabox, line 25:
```suggestion
# Note this is moderately evil accessing julia's internals
@generated function _force_construct(T, args...)
```

oxinabox (approved):

PR #183, opened 2021-07-21
deprecate rand_tangent

oxinabox:
Partner to https://github.com/JuliaDiff/ChainRulesTestUtils.jl/pull/189

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -1,3 +1,10 @@
+function depwarn_rt()
+    Base.depwarn(
+        "FiniteDifferences.rand_tangent is deprecated, it has moved to ChainRulesTestUtils",
+        :rand_tangent
+    )
+end
+
 """
     rand_tangent([rng::AbstractRNG,] x)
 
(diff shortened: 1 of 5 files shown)


src/deprecated.jl
```diff
@@ -53,11 +66,12 @@ function rand_tangent(rng::AbstractRNG, x::T) where {T}
     end
     if all(tangent isa NoTangent for tangent in tangents)
         # if none of my fields can be perturbed then I can't be perturbed
+        depwarn_rt()
```
mzgubic, line 69: should this be outside of the if statement so that both branches get it?

test/deprecated.jl
```diff
@@ -1,6 +1,12 @@
-using FiniteDifferences: rand_tangent
+# Test struct for `rand_tangent` and `difference`.
+struct Foo
+    a::Float64
+    b::Int
+    c::Any
+ end
+rand_tangent(args...) = @test_deprecated FiniteDifferences.rand_tangent(args...)
```
mzgubic, line 7: that's pretty cool - you are defining a new function here, which calls the original one?

mzgubic (approved):
version bump and good to go

test/deprecated.jl
oxinabox: Yeah, It is pretty lazy.

```suggestion

# to avoid deprecation spam (and actually test deprecations) we will define a wrapper `rand_tangent` function for testing
rand_tangent(args...) = @test_deprecated FiniteDifferences.rand_tangent(args...)
```

src/deprecated.jl
oxinabox: No, the other branch calls `rand_tangent`, so will git it somewhere else.
Having it outside the if would result in it being printed multiple times

Issue #186, opened 2021-07-27
`to_vec(::Diagonal)`: densify or not?


mzgubic:
This issue is just putting down some of my notes on why I think the current status is ok.

The question is whether `to_vec(::Diagonal)` should return only the vector along the diagonal, or the whole densified matrix. I.e. should it do
```julia
julia> v, b = to_vec(Diagonal([1, 2])); v # struct-like
2-element Vector{Int64}:
 1
 2
```
or
```julia
julia> v, b = to_vec(Diagonal([1, 2])); v # array-like
4-element Vector{Int64}:
 1
 0
 0
 2
```
in other words, do we treat it as a struct or as an array? At present, we densify it, which feels wrong in the sense that the off-diagonal zeros are structural, i.e. they can't (conceptually) be perturbed. Practically, when we perturb the off-diagonal elements, we just get no effect and we waste computation.

On the other hand, it is practical. Consider `test_rrule(Symmetric, randn(3, 3), :U)`, and we want to test whether the rule accepts `Diagonal` tangents as well, which we do by `test_rrule(Symmetric, randn(3, 3), :U; check_inferred=false, output_tangent=Diagonal(rand(3, 3)))`. If we do not densify the Diagonal, the test fails. The reason is that the finite differences jacobian is computed for the output type (Symmetric) vector (length 9), while the non-densified vector for a 3by3 Diagonal has length 3, which causes an awkward `DimensionMismatch("second dimension of A, 9, does not match length of x, 3")` error, which is highly likely confusing to a newcomer.

So the tradeoff seems to be sacrificing convenience or wasting computational power. In the 21st century convenience is sacred and computation is cheap, so keeping Diagonal densified is ok with me.

---------

Densifying used to cause different (and quite common) problems when testing rules in ChainRulesTestUtils, e.g. for `d=Diagonal(rand(3)); test_rrule(*, d, d)` because the `rand_tangent` would make a `Tangent{Diagonal}(diag=[1., 2., 3.])` (vector of length 3), whereas the primal was a `Diagonal` (vector of length 9). However these have been mostly fixed by adding the `ProjectTo` into `rand_tangent` in ChainRulesCore 1.0.

PR #187, opened 2021-08-01
WIP: Change matrix to_vec semantics

willtebbutt:
Relates to / resolves https://github.com/JuliaDiff/FiniteDifferences.jl/issues/132

Since `ChainRulesTestUtils` no longer relies on checking equality in the `to_vec`ed space (I don't believe), we're freed up to make this change.

In short: after this change, it will generally be a really bad idea to check equality in the `to_vec`ed space (I _think_ it was probably a bad idea anyway to be honest, because it's generally a lossy transformation).

The motivation for making this PR now was that I was trying to implement some `rrule`s, and encountered the issue that `to_vec(rand_tangent(::Diagonal))` and `to_vec(::Diagonal)` yield different sized vectors, which causes problems with `FiniteDifferences`'s `jpvp` and / or `jvp` functionality.

ToDo:
- [ ] check that this doesn't break ChainRules

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.12.18"
+version = "0.13.0"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -18,6 +18,12 @@ function to_vec(z::Complex)
     return [real(z), imag(z)], Complex_from_vec
 end
 
+# Integers cannot be perturbed!
+function to_vec(x::Integer)
+    Integer_from_vec(v) = x
+    return Bool[], Integer_from_vec
+end
+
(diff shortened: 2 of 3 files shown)


mzgubic:
This will very much break ChainRules unfortunately :(. I tried this recently and wrote something up here for diagonal: https://github.com/JuliaDiff/FiniteDifferences.jl/issues/186

Also, `rand_tangent(::Diagonal)` now returns a `Diagonal` (not a `Tangent{Diagonal}`) since `rand_tangent` has been moved back to ChainRulesTestUtils and `ProjectTo` used.

I do think though that we are missing something in our finite differencing story. One aspect is https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/199 for example. It looks like we need something similar to `ProjectTo`

willtebbutt:
Thanks for pointing out that you've already looked at this @mzgubic -- turns out I've been using an outdated ChainRulesTestUtils version anyway...

I'll close for now. I agree that there's other stuff to sort out.

mzgubic:
Oh, I see - I found that the newer ChainRulesTestUtils version with `project` inside `rand_tangent` for structured arrays helps solve quite a few of the more annoying things we've been running into

PR #189, opened 2021-08-04
make `to_vec(::Integer)` an empty vector

### LINKED ISSUE #188: Should `to_vec(::Integer)` return an empty vector
Integers can't be perturbed, so their `to_vec` should be empty. Originally suggested in https://github.com/JuliaDiff/FiniteDifferences.jl/pull/187
```julia
function to_vec(x::Integer)
     Integer_from_vec(v) = x
     return Bool[], Integer_from_vec
end
```

Would also fix (a part of) https://github.com/chrisbrahms/Hankel.jl/pull/27

Did not test yet whether it messes up any existing ChainRules tests.

mzgubic:
Closes #188 

breaks a few rules in ChainRules, need to take a look at that before merging

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.12.18"
+version = "0.12.19"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -54,7 +54,7 @@ is defined. Each 2-`Tuple` in `xẋs` contains the value `x` and its tangent `x�
 """
 function jvp(fdm, f, (x, ẋ)::Tuple{Any, Any})
     x_vec, vec_to_x = to_vec(x)
-    y, vec_to_y = to_vec(f(x))
+    _, vec_to_y = to_vec(f(x))
     return _int2zero(vec_to_y(_jvp(fdm, x_vec->to_vec(f(vec_to_x(x_vec)))[1], x_vec, to_vec(ẋ)[1])))
 end
 function jvp(fdm, f, xẋs::Tuple{Any, Any}...)
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -54,8 +54,8 @@ is defined. Each 2-`Tuple` in `xẋs` contains the value `x` and its tangent `x�
 """
 function jvp(fdm, f, (x, ẋ)::Tuple{Any, Any})
     x_vec, vec_to_x = to_vec(x)
-    _, vec_to_y = to_vec(f(x))
-    return vec_to_y(_jvp(fdm, x_vec->to_vec(f(vec_to_x(x_vec)))[1], x_vec, to_vec(ẋ)[1]))
+    y, vec_to_y = to_vec(f(x))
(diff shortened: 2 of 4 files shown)


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.12.18"
+version = "0.12.19"
```
mzgubic, line 3: should this be breaking instead? i think of it as a bug fix

test/to_vec.jl
```diff
@@ -67,8 +67,11 @@ function test_to_vec(x::T; check_inferred=true) where {T}
     return nothing
 end
 
+myrandn(T::Type{<:Number}, args...) = randn(T, args...)
```
mzgubic, line 70: i should probably rename this to something sensible.

`robust_randn`, `randn_ints_ok`, `randn_that_doesnt_hate_ints`

Project.toml
willtebbutt: My view is treat it as a bug fix for now -- if it turns out to be super breaking and we later decide it's actually a breaking change, we can always bump the minor version number then.

src/grad.jl
```diff
@@ -85,4 +85,14 @@ end
...
 """
-grad(fdm, f, xs...) = j′vp(fdm, f, 1, xs...)  # `j′vp` with seed of 1
+grad(fdm, f, xs...) = j′vp(fdm, f, 1.0, xs...)  # `j′vp` with seed of 1
+
+# This deals with the fact that integers are non perturbable
+# v, b = to_vec(1);
+# v == []
+# b(v) == 1
+# which means that jvp always returns the integer itself, since [] - [] == []
+_int2zero(x) = x
```
willtebbutt, line 95: I suspect that this will miss a few cases (e.g. vectors of integers), but probably that's fine for now -- we can revisit later it needs be.

willtebbutt: On a more general note: is the need for this function a manifestation of the fact that FiniteDifferences doesn't really know how to handle tangents properly?

test/to_vec.jl
willtebbutt: I'm not sure this should have `randn` in the name, since the integer method doesn't really sample from a normal distribution. Perhaps something a bit more verbose like `rand_number`?

mzgubic:
I tackled this before leaving for holidays but didn't have time to continue when I came back. I do remember it was kind of annoying as some weird things were popping up (one symptom is the `_int2zero` bandaid)

willtebbutt:
I _think_ the underlying problem is related to Example 2 from [this CRTU issue](https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/213). Need to confirm by working through the things that are breaking though.

willtebbutt:
> I think the underlying problem is related to Example 2 from this CRTU issue. Need to confirm by working through the things that are breaking though.

I did a bit more digging into what it would take to implement this. It turns out that we're also missing a way to say "treat `t` as a tangent for `x` and convert it to a vector of reals". This functionality is needed in `j′vp` The distinction is most plain for integers:

`to_vec(5)` should clearly return an empty vector because integers are non-differentiable, however, if we have a function `tangent_to_vec(primal, tangent)` where `tangent` is a tangent for `primal`, `tangent_to_vec(1.0, 5)` should return a length 1 vector containing `5`.

Similarly, while `to_vec(primal::Diagonal)` should return a length `size(primal, 1)` vector (if we're treating things structurally), `tangent_to_vec(primal::Matrix{Float64}, tangent::Diagonal{Float64})` should return a length `length(primal)` vector.

Very doable, just requires some work.

PR #191, opened 2021-09-27
Fix to_vec with Integers

willtebbutt:
Integers aren't considered differentiable, and they're certainly not perturbable. This bug fix enforces that.

This will probably break some things downstream that depend on this bug.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.12.18"
+version = "0.12.19"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/grad.jl
+++ b/src/grad.jl
@@ -85,4 +85,4 @@ end
 
 Compute the gradient of `f` for any `xs` for which [`to_vec`](@ref) is defined.
 """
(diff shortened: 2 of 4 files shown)


oxinabox:
What was the previous behavour?

mzgubic:
>What was the previous behavour?

I think previously it just treated the integers as embedded in the reals and returned a vector of length 1 with integer transformed to a real.

>This will probably break some things downstream that depend on this bug.

We might want to add tests for this as well, like in:
https://github.com/JuliaDiff/FiniteDifferences.jl/pull/189/files

Also, the `_int2zero` in that PR was to get around some really annoying behaviour where `j'vp` is not computed properly. The reason is the following situation:

```julia
julia> v1, b1 = to_vec(6)

julia> v2, b2 = to_vec(6)

julia> res = b1(v1 - v2) # b1 will just return 6 whatever the vector is
6
```

`_int2zero` didn't feel like the right solution at the time, and it still doesn't. Maybe there is a better way around this

willtebbutt:
> I think previously it just treated the integers as embedded in the reals and returned a vector of length 1 with integer transformed to a real.

Exactly.

> We might want to add tests for this as well, like in:
https://github.com/JuliaDiff/FiniteDifferences.jl/pull/189/files

Lol sorry, I totally forgot about that PR. I'll close this PR in favour of that.

mzgubic:
I am not working on it actively, feel free to take it over!

PR #202, opened 2022-02-11
implement `to_vec` for sparse arrays

mzgubic:
Needed for testing https://github.com/JuliaDiff/ChainRules.jl/pull/586

### DIFF
--- a/src/to_vec.jl
+++ b/src/to_vec.jl
@@ -156,6 +156,30 @@ function to_vec(X::T) where {T<:PermutedDimsArray}
     return x_vec, PermutedDimsArray_from_vec
 end
 
+function to_vec(v::SparseVector)
+    inds, values = findnz(v)
+    sizes = size(v)
+
+    x_vec, back = to_vec(values)
+    function SparseVector_from_vec(x_v)
+        v_values = back(x_v)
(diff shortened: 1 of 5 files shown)


src/to_vec.jl
```diff
@@ -156,6 +156,30 @@ function to_vec(X::T) where {T<:PermutedDimsArray}
...
+        return sparsevec(inds, v_values[inds], sizes...)
+    end
+    return x_vec, SparseVector_from_vec
+end
+
+function to_vec(m::SparseMatrixCSC)
+    is, js, _ = findnz(m)
+    sizes = size(m)
+
+    x_vec, back = to_vec(collect(m))
```
willtebbutt, line 175: Why is collecting the right thing to do here, rather than treating these types like structs?

mzgubic: This is the use case: https://github.com/JuliaDiff/ChainRules.jl/pull/586/files and both solutions work for that. I've chosen this one to be consistent with what we do with e.g. Diagonal matrix, to avoid issues described in https://github.com/JuliaDiff/ChainRulesTestUtils.jl/issues/213

willtebbutt (approved):
LGTM

PR #212, opened 2023-01-08
Matrixify Q's in tests

dkarrasch:
The comparison involves elementwise `getindex` anyway, so it should be faster to first compute the matrix and then compare. Also, this will stop working after https://github.com/JuliaLang/julia/pull/46196 is merged.


dkarrasch:
I tested this locally on my PR branch. Tests pass, so this is ready for review/merge/release.

PR #213, opened 2023-01-09
Revert toml bump

willtebbutt:
#212 bumped the patch version, but I don't think it needed to since there were no changes to src, and therefore nothing to release (this is discussed somewhere in colprac or something).

@wesselb @dkarrasch are you happy for this reversion to go through and a release to not be made, or is there a reason that I'm not aware of for why a patch release should be made?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.12.26"
+version = "0.12.25"
 
 [deps]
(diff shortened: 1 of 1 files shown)


wesselb (approved):
Thanks @willtebbutt! No, please do revert. I didn’t catch this. :)

dkarrasch:
Sorry, I forgot to explain. I'd like this to be released so that PkgEval picks this up and doesn't give "spurious" fails in https://github.com/JuliaLang/julia/pull/46196. But it's perhaps not too important, so do as you prefer.

willtebbutt:
Thanks for the explanation @dkarrasch -- seems reasonable to me. Will close this and tag a release.

PR #225, opened 2023-07-13
Fix overflow catch in coeffs function

alexfreudenberg:
The current implementation doesn't seem to catch integer overflows correctly. Consider the following:

```{julia}
julia> 2^62
4611686018427387904

julia> 2^63
-9223372036854775808

julia> Rational{BigInt}(2^62*2)
-9223372036854775808//1

julia> Rational{BigInt}(2^62)*2
9223372036854775808//1

```


wesselb:
Yes, very good catch, @alexfreudenberg!! Thank you. :) I think this can just be merged right away!

PR #244, opened 2025-08-28
Add Unitful support

### LINKED ISSUE #243: Accept Unitful numbers as input
Currently, when attempting to use unitful numbers from [Unitful.jl](https://github.com/JuliaPhysics/Unitful.jl) with the functions in this library, one gets `MethodErrors` related to the fact that `Unitful.Quantity{...} <: Number` instead of `Unitful.Quantity{...} <: Real`:
```sh
ERROR: MethodError: no method matching (::FiniteDifferences.AdaptedFiniteDifferenceMethod{…})(::var"#35#36", ::Quantity{…})
The object of type `FiniteDifferences.AdaptedFiniteDifferenceMethod{3, 1, FiniteDifferences.UnadaptedFiniteDifferenceMethod{5, 3}}` exists, but no method is defined for this combination of argument types when trying to treat it as a callable object.

Closest candidates are:
  (::FiniteDifferences.AdaptedFiniteDifferenceMethod)(::TF, ::Real) where TF
   @ FiniteDifferences ~/.julia/packages/FiniteDifferences/IPGFN/src/methods.jl:191
  (::FiniteDifferences.AdaptedFiniteDifferenceMethod{P, Q})(::TF, ::Real, ::Real) where {P, Q, TF}
   @ FiniteDifferences ~/.julia/packages/FiniteDifferences/IPGFN/src/methods.jl:238
```
At this point I am not sure whether simply relaxing the argument type restriction of functions like `FiniteDifferences.central_fdm` from `Real` to `Number` would suffice (with the possible risk of introducing silent semantic errors), or whether separate methods, that explicitly handle the units in some way (e.g. first stripping them and then reattaching the proper ones to the results) should be defined for Unitful numbers.

Regardless, it would be *very* nice if this library supported unitful numbers out of the box.

SeSodesa:
Makes the functions in `src/methods` accept `Numbers` instead of just `Real`s of `AbstractFloats`. This makes them almost compatible with `Quantity <: Number` in Unitful.jl, however units also need to be handled in the function implementations. Towards this end, "dimensional analysis" is performed inside of functions like `estimate_step`.

Closes #243.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -10,6 +10,7 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Richardson = "708f8203-808e-40c0-ba2d-98a6953ed40d"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
+Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
 ChainRulesCore = "0.10.3, 1"
@@ -19,6 +20,7 @@ Random = "<0.0.1, 1"
 Richardson = "1.2"
 SparseArrays = "<0.0.1, 1"
 StaticArrays = "0.12, 1.0"
+Unitful = "1.24.0"
 julia = "1"
 
 [extras]
--- a/src/FiniteDifferences.jl
+++ b/src/FiniteDifferences.jl
@@ -7,6 +7,7 @@ using Random
 using Richardson
 using SparseArrays
 using StaticArrays
+using Unitful
 
 export to_vec, grad, jacobian, jvp, j′vp
 
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -261,7 +261,8 @@ function _compute_estimate(
     # https://github.com/JuliaLang/julia/issues/39151.
     #
     # We strip units because the estimate coefficients are just weights for values of f.
-    _coefs = ustrip.(T.(coefs))
+    N = numType(T)
+    _coefs = N.(coefs)
     return sum(fs .* _coefs) ./ T(step) ^ Q
 end
 
@@ -358,27 +359,26 @@ estimate of the derivative.
 function estimate_step(
     m::UnadaptedFiniteDifferenceMethod, f::TF, x::T,
 ) where {TF,T<:Number}
-    step, acc = withUnit.(
-        unit(x),
-        _compute_step_acc_default(m, x)
-    )
+    step, acc = _compute_step_acc_default(m, x)
+    xunit = unit(x)
+    step = withUnit(xunit, step)
+    acc = withUnit(xunit,acc)
     return _limit_step(m, x, step, acc)
 end
 function estimate_step(
     m::AdaptedFiniteDifferenceMethod{P,Q}, f::TF, x::T,
 ) where {P,Q,TF,T<:Number}
     ∇f_magnitude, f_magnitude = _estimate_magnitudes(m.bound_estimator, f, x)
-    step, acc = withUnit.(
-        (
-            unit(x),
-            unit(first(f(x))) / unit(x) ^ Q
-        ),
+    xunit = unit(x)
+    dfunit = unit(first(f(x))) / unit(x) ^ Q
+    step, acc =
         if ∇f_magnitude == withUnit(unit(∇f_magnitude),0.0) || f_magnitude == withUnit(unit(f_magnitude), 0.0)
             _compute_step_acc_default(m, x)
         else
             _compute_step_acc(m, ∇f_magnitude, eps(f_magnitude))
         end
-    )
+    step = withUnit(xunit, step)
+    acc = withUnit(dfunit, acc)
     return _limit_step(m, x, step, acc)
 end
 
@@ -431,10 +431,8 @@ function _limit_step(
     end
     # Second, prevent very large step sizes, which can occur for high-order methods or
     # slowly-varying functions.
-    step_default, _ = withUnit.(
-        xunit,
-        _compute_step_acc_default(m, x)
-    )
+    step_default, _ = _compute_step_acc_default(m, x)
+    step_default = withUnit(xunit, step_default)
     step_max_default = 1000step_default
     if step > step_max_default
         step = step_max_default
@@ -628,3 +626,22 @@ function withUnit(targetUnit, value)
     end # if
 
 end # function
+
+"""
+Retrieves the number type of a quantity, or returns the type itself in the case of a raw number.
+"""
+function numType(x::Number)
+    typeof(x)
+end # function
+
+function numType(x::Type{<:Number})
+    x
+end
+
+function numType(x::Unitful.AbstractQuantity)
+    Unitful.numtype(typeof(x))
+end # function
+
+function numType(x::Type{<:Unitful.AbstractQuantity})
+    Unitful.numtype(x)
+end
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -261,7 +261,7 @@ function _compute_estimate(
     # https://github.com/JuliaLang/julia/issues/39151.
     #
     # We strip units because the estimate coefficients are just weights for values of f.
-    _coefs = ustrip(T.(coefs))
+    _coefs = ustrip.(T.(coefs))
     return sum(fs .* _coefs) ./ T(step) ^ Q
 end
 
@@ -360,7 +360,7 @@ function estimate_step(
 ) where {TF,T<:Number}
     step, acc = withUnit.(
         unit(x),
-        _compute_step_acc_default(m, x) .* unit(x)
+        _compute_step_acc_default(m, x)
     )
     return _limit_step(m, x, step, acc)
 end
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -371,9 +371,9 @@ function estimate_step(
     step, acc = withUnit.(
         (
             unit(x),
-            unit(f |> eltype) / unit(x) ^ Q
+            unit(first(f(x))) / unit(x) ^ Q
         ),
-        if ∇f_magnitude == withUnit(∇f_magnitude,0.0) || f_magnitude == withUnit(unit(f_magnitude), 0.0)
+        if ∇f_magnitude == withUnit(unit(∇f_magnitude),0.0) || f_magnitude == withUnit(unit(f_magnitude), 0.0)
             _compute_step_acc_default(m, x)
         else
             _compute_step_acc(m, ∇f_magnitude, eps(f_magnitude))
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -367,7 +367,7 @@ end
 function estimate_step(
     m::AdaptedFiniteDifferenceMethod{P,Q}, f::TF, x::T,
 ) where {P,Q,TF,T<:Number}
-    @show ∇f_magnitude, f_magnitude = _estimate_magnitudes(m.bound_estimator, f, x)
+    ∇f_magnitude, f_magnitude = _estimate_magnitudes(m.bound_estimator, f, x)
     step, acc = withUnit.(
         (
             unit(x),
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -107,9 +107,9 @@ end
     FiniteDifferenceMethod(
         grid::AbstractVector{Int},
         q::Int;
-        condition::Real=DEFAULT_CONDITION,
-        factor::Real=DEFAULT_FACTOR,
-        max_range::Real=Inf
+        condition::Number=DEFAULT_CONDITION,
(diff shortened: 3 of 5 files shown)


SeSodesa:
Note that this might rely on Unitful.jl merging <https://github.com/JuliaPhysics/Unitful.jl/pull/807>, although that is not quaranteed, since the propagation of the `unit` call to the method `unit(::Any)` was probably caused by me originally assuming, that `eltype` was somehow valid for objects `f::FT where FT` passed into the different functions in the library. It was not.

**Edit:** nope, the current release of Unitful also works.

oxinabox:
The relaxation of types constraints is fine.
I am less keen on the extra complexity of the code that removes units then adds them back etc.

I would at least like to see benchmarks showing that they all compile away fully for the case of normal `Float64` and `Vector{Float64}` inputs and that our performance does not change with/without this

SeSodesa:
The benchmarks are definitely still a concern of mine. My guestion is, what would constitute a good benchmark that would satisfy most people? I guess I could `rg` the test folder for the words "allocation" or "benchmark", and then write a unitful version of the related tests?

SeSodesa:
As a quick note, when called with numbers without units, the allocations seem to remain small, while unitful calls currently increase the number of allocations by some amount:
```
julia> using Unitful, FiniteDifferences, BenchmarkTools

julia> fn(t) = sin(t / u"s")
fn (generic function with 1 method)

julia> central_fdm(5,1)(fn, pi * u"s")
-1.000000000000011 s⁻¹

julia> @benchmark central_fdm(5,1)(fn, pi * u"s")
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  15.500 μs … 152.375 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     16.458 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   16.848 μs ±   2.209 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

    ▅██▆▄ ▃▂
  ▂▆██████████▆▅▄▅▆▆▅▅▄▄▃▃▃▃▄▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂ ▃
  15.5 μs         Histogram: frequency by time         22.7 μs <

 Memory estimate: 13.04 KiB, allocs estimate: 262.

julia> @benchmark central_fdm(5,1)(sin, pi)
BenchmarkTools.Trial: 10000 samples with 10 evaluations per sample.
 Range (min … max):  1.837 μs …   5.842 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     1.958 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.988 μs ± 135.514 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

      ▄▃▃▅▆█▃
  ▂▃▅▇████████▇█▇▆▅▅▄▄▃▃▃▃▂▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂ ▃
  1.84 μs         Histogram: frequency by time        2.63 μs <

 Memory estimate: 2.45 KiB, allocs estimate: 35.
```
I'll have to check whether the allocations could be avoided in the Unitful case, even if they are not increased for the unitless one. Also, note the incorrect units of the derivative of a unitless function `fn`. That should still also be taken into account.

**Edit:** nevermind, the units are correct. Working on multiple things at the same time again...

SeSodesa:
The easiest thing might be to simply record what the units should be at the start based on the order of the derivative, and then just adding them back in at the end.

oxinabox:
`sin` is fine as a thing to benchmark.


 I would benchmark the following FDM methods:
 
 ```
 central_fdm(5, 1)  # default adapt is 1
 central_fdm(5, 1; max_range=1e-2)  # this is what chainrules uses in its tests
 central_fdm(5, 1; adapt=0)
 central_fdm(5, 1; adapt=2)
 ```
 because those are the critical ones that people use,
 and if they slow down it's bad.
 Especially for chain rules which has thousands of tests

wesselb:
> The easiest thing might be to simply record what the units should be at the start based on the order of the derivative, and then just adding them back in at the end.

I think this is a good way to go about it. Once the units of the function value and inputs and the order of the derivative is know, the units of the output can be computed.

SeSodesa:
Ah, but the problem with only assigning the units at the end is that unit information might be embedded into the function whose derivative is being estimated, as was done above:
```julia
fn(t) = sin(t / u"t")
```
For the purposes of computing the numerical value of a derivative first and then attaching the units, we can strip unit information from the number `x` at which the derivative is being approximated, but as far as I know, there is no way to strip the unit information from within a function. This is why units have to be handed by `estimate_step` and the functions that it calls.

Now, I am still a bit baffled as to why there are more allocations in the current code than in the purely numerical one. Attaching and stripping units from scalars alone should not cause allocations, and there should be no dotted function calls in the WIP code anymore. Maybe it has something to do with type instability, since this is the output I get when adding a few debug prints to the code:
```julia
julia> cdm = central_fdm(5,1)
FiniteDifferenceMethod:
  order of method:       5
  order of derivative:   1
  grid:                  [-2, -1, 0, 1, 2]
  coefficients:          [0.08333333333333333, -0.6666666666666666, 0.0, 0.6666666666666666, -0.08333333333333333]


julia> cdm(t -> sin(t/u"s"),pi * u"s")

1
FiniteDifferences.AdaptedFiniteDifferenceMethod = FiniteDifferences.AdaptedFiniteDifferenceMethod
typeof(m) = FiniteDifferences.AdaptedFiniteDifferenceMethod{5, 1, FiniteDifferences.UnadaptedFiniteDifferenceMethod{7, 5}}
x = π s
f = var"#85#86"()

estimate_step
x = 3.141592653589793 s
f(x) = 1.2246467991473532e-16

estimate_magnitudes
x = 3.141592653589793 s
f(x) = 1.2246467991473532e-16

_compute_step_acc_default
x = 3.141592653589793 s

_compute_step_acc
∇f_magnitude = 10.0
f_error = 2.220446049250313e-16

_limit_step
x = 3.141592653589793 s
step = 0.007231988488633704 s

_compute_step_acc_default
x = 3.141592653589793 s

_compute_step_acc
∇f_magnitude = 10.0
f_error = 2.220446049250313e-16
x = 3.141592653589793 s
f(x) = 1.2246467991473532e-16

_compute_estimate
x = 3.141592653589793 s
fs = [0.021694263403528785, 0.014463472655896089, 0.007231925447967751, 1.2246467991473532e-16, -0.007231925447967506, -0.014463472655895844, -0.02169426340352854]

_compute_estimate
x = 3.141592653589793 s
fs = [0.021694263403528785, 0.014463472655896089, 0.007231925447967751, 1.2246467991473532e-16, -0.007231925447967506, -0.014463472655895844, -0.02169426340352854]

_compute_estimate
x = 3.141592653589793 s
fs = [0.021694263403528785, 0.014463472655896089, 0.007231925447967751, 1.2246467991473532e-16, -0.007231925447967506, -0.014463472655895844, -0.02169426340352854]

_compute_step_acc
∇f_magnitude = 1.0000053814049705 s^-5
f_error = 3.469446951953614e-18

_limit_step
x = 3.141592653589793 s
step = 0.0004719677647195769 s

_compute_step_acc_default
x = 3.141592653589793 s

_compute_step_acc
∇f_magnitude = 10.0
f_error = 2.220446049250313e-16

1
FiniteDifferences.AdaptedFiniteDifferenceMethod = FiniteDifferences.AdaptedFiniteDifferenceMethod
step = 0.0004719677647195769 s

2
FiniteDifferences.AdaptedFiniteDifferenceMethod = FiniteDifferences.AdaptedFiniteDifferenceMethod
x = 3.141592653589793 s
f(x) = 1.2246467991473532e-16

_compute_estimate
x = 3.141592653589793 s
fs = [0.0009439353892626215, 0.00047196774719762154, 1.2246467991473532e-16, -0.00047196774719737657, -0.0009439353892623766]
-1.000000000000011 s^-1
```
So initially `∇f_magnitude` does not have units, but they get attached to it after a few calls of `_compute_step_acc`, and then stripped again.

wesselb:
> (…) there is no way to strip the unit information from within a function.

For a given function `f` which might take in inputs with units and/or produce outputs with units, would it be an idea to (1) wrap `f` into `f_wrapped` where `f_wrapped` is exactly like `f` but without any units on the inputs/outputs, (2) call the finite-difference method on `f_wrapped`, and (3) return the derivative with the appropriate units?

SeSodesa:
I think the problem there is that if `f_wrapper` was defined as something like a closure
```julia
# The function f is passed to fdm and captured.
f_wrapper(x) = x -> ustrip(f(x))
```
where `f` comes from the surrounding context, the internal functions that call `f` do so in a dotted manner in `_eval_function`, so this might result in an allocation. Or maybe I am imagining things.

SeSodesa:
Ok, it might not be an issue:
```julia
julia> using Revise, Unitful, StaticArrays, BenchmarkTools, FiniteDifferences

julia> fn(t) = sin(t / u"s")
fn (generic function with 1 method)

julia> fnwrap(x) = ustrip(fn(x))
fnwrap (generic function with 1 method)

julia> @benchmark fnwrap(1.0u"s")
BenchmarkTools.Trial: 10000 samples with 1000 evaluations per sample.
 Range (min … max):  0.833 ns … 3.750 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     0.875 ns             ┊ GC (median):    0.00%
 Time  (mean ± σ):   0.895 ns ± 0.062 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

       █     ▇                           ▁                  ▁
  █▁▁▁▁█▁▁▁▁▁█▁▁▁▁▁▇▁▁▁▁█▁▁▁▁▁▇▁▁▁▁▁▆▁▁▁▁█▁▁▁▁▁▅▁▁▁▁▁▄▁▁▁▁█ █
  0.833 ns    Histogram: log(frequency) by time     1.25 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark fnwrap.((1.0u"s", 2u"s"))
BenchmarkTools.Trial: 10000 samples with 1000 evaluations per sample.
 Range (min … max):  0.833 ns … 20.000 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     0.875 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   0.890 ns ±  0.194 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                     █                  ▆                    ▁
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇ █
  0.833 ns     Histogram: log(frequency) by time    0.959 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark fnwrap.(SVector((1.0u"s", 2u"s")))
BenchmarkTools.Trial: 10000 samples with 1000 evaluations per sample.
 Range (min … max):  0.833 ns … 8.625 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     0.875 ns             ┊ GC (median):    0.00%
 Time  (mean ± σ):   0.892 ns ± 0.097 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

          █       ▆                                         ▁
  █▁▁▁▁▁▁▁█▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▇▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▆▁▁▁▁▁▁▁▆▁▁▁▁▁▁▁▆ █
  0.833 ns    Histogram: log(frequency) by time     1.12 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark fnwrap.([1.0u"s", 2.0u"s"])
BenchmarkTools.Trial: 10000 samples with 998 evaluations per sample.
 Range (min … max):  18.996 ns …  1.018 μs  ┊ GC (min … max):  0.00% … 97.05%
 Time  (median):     19.998 ns              ┊ GC (median):     0.00%
 Time  (mean ± σ):   23.651 ns ± 36.816 ns  ┊ GC (mean ± σ):  12.65% ±  7.91%

   ▃  █▂        ▄
  ▇█████▄▃▃▄▃▃▃▅██▅▇▇▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▁▂▂ ▃
  19 ns           Histogram: frequency by time          29 ns <

 Memory estimate: 160 bytes, allocs estimate: 4.
```

wesselb:
That looks promising!

PR #245, opened 2025-09-12
Fix step size calculation if some results are non-deterministic (inf …

wsmoses:
…derivative) or nan

step size is calculated based of the initial derivatives of all results. However, if some of the results return nan and/or are non-determinstic, the step size will be inf/nan, causing all other results to be broken -- instead of just the result which is expected to have an inf/nan derivative

cc @vchuravy @giordano

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -391,6 +391,7 @@ function _estimate_magnitudes(
         _compute_estimate(m, fs, x, step, m.coefs_neighbourhood[2]),
         _compute_estimate(m, fs, x, step, m.coefs_neighbourhood[3])
     )
+    ∇fs = finite_or_zero(∇fs)
     ∇f_magnitude = maximum(maximum.(abs, ∇fs))
     # Estimate magnitude of `f` in a neighbourhood of `x`.
     f_magnitude = maximum(maximum.(abs, fs))
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -371,11 +371,20 @@ function estimate_step(
     return _limit_step(m, x, step, acc)
 end
 
+function finite_or_zero(fs::AbstractArray{<:Number})
+   ifelse.(isfinite.(fs), fs, zero(fs))
+end
+
+function finite_or_zero(fs::AbstractArray{<:AbstractArray})
+   finite_or_zero.(fs)
+end
+
 function _estimate_magnitudes(
     m::FiniteDifferenceMethod{P,Q}, f::TF, x::T,
 ) where {P,Q,TF,T<:AbstractFloat}
     step = first(estimate_step(m, f, x))
     fs = _eval_function(m, f, x, step)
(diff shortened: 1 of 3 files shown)


test/grad.jl
```diff
@@ -217,3 +217,33 @@ using FiniteDifferences: grad, jacobian, _jvp, jvp, j′vp, _j′vp, to_vec
         @test [real(ȳ), imag(ȳ)] ≈ Jy'z̄_vec
     end
 end
+
+using LinearAlgebra
+
+function partial_nan_returning(x)
```
oscardssmith, line 223: now that we know the issue, can't we use a simpler mwe for the test? i.e. a JVP with non-random data

oscardssmith (changes requested):

test/grad.jl
wsmoses: I don't follow?

wsmoses: a Nan returning function does explicilty need checking and would result in the failure (and fairly simply)

as does any non-deterministic (and specifically different in each call) output. This is fairly simple imo and explicitly tests both of these (while not relying on nondeterminsim from partially initialized arrrays)

src/methods.jl
```diff
@@ -371,11 +371,20 @@ function estimate_step(
...
+function finite_or_zero(fs::AbstractArray{<:AbstractArray})
+   finite_or_zero.(fs)
+end
+
 function _estimate_magnitudes(
     m::FiniteDifferenceMethod{P,Q}, f::TF, x::T,
 ) where {P,Q,TF,T<:AbstractFloat}
     step = first(estimate_step(m, f, x))
     fs = _eval_function(m, f, x, step)
+    fs = finite_or_zero(fs)
```
giordano, line 387: Can fuse with previous line, instead of defining fs twice (I think the compiler can fuse nested broadcasts across function calls? but of course not if the calls are separated)

oscardssmith: broadcast is syntactic. lines don't make a difference

wsmoses:
relatedly @oscardssmith can we add EnzymeTestUtils as reverse CI here too [as I see ChainRulesTestUtils as such a a reverse CI]

test/grad.jl
wsmoses: done per slack disc

oscardssmith:
I would defer on CI decisions to someone more deeply involved in this package.

oscardssmith (approved):

wsmoses:
fair and yeah can follow up later on that.

ci chainrules error is unrelated

oscardssmith:
I'm not aware of who actually has merge rights here. Looks like the most recent PRs merged were last year by @oxinabox, and @wesselb, so presumably this should get a review from one of them.

oxinabox (approved):

PR #247, opened 2025-09-15
Update badges and CI

devmotion:
A bit of housekeeping: This PR fixes broken and outdated badges and updates CI (eg testing of `min` instead of hardcoded `1.0`; removing hardcoded architectures which are not appropriate for Apple silicon; adding an SSH key to CompatHelper; using `julia-actions/cache` instead of manually configured `actions/cache`).


.github/workflows/Documenter.yml
```diff
@@ -10,14 +10,13 @@ jobs:
     name: Documentation
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - uses: julia-actions/setup-julia@v2
         with:
-          version: '1.10'
+          version: '1'
```
oxinabox, line 16: locking it to a precise minor version seems good.
Because it stops doctests breaking if julia decides to change the string representation of things again

```suggestion
          version: '1.10'
```

.gitignore
```diff
@@ -1,7 +1,7 @@
 *.jl.cov
 *.jl.*.cov
 *.jl.mem
-/Manifest.toml
+Manifest.toml
```
oxinabox, line 4: keeping the docs manifest around is useful to avoid breaking doctests if packages we use change their string represetations

oxinabox (approved):

.github/workflows/Documenter.yml
devmotion: 1.10 doesn't support `path`, IIRC at least 1.11 is needed.

.github/workflows/Documenter.yml
```diff
@@ -10,14 +10,13 @@ jobs:
     name: Documentation
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - uses: julia-actions/setup-julia@v2
         with:
           version: '1.10'
```
devmotion:
```suggestion
          version: '1'
```

.gitignore
devmotion: IMO such doctest failures should be avoided by using doctest filters or changing to more stable tests, since string representations may change even in non-breaking releases.

.github/workflows/CI.yml
```diff
@@ -7,43 +7,37 @@ on:
...
-        # 32-bit Julia binaries are not available on macOS
         exclude:
-          - os: macOS-latest
-            arch: x86
+          # aarch64 binaries are not available for Julia 1.0
+          - version: 'min'
+            os: 'macOS-latest'
+        include:
+          - version: 'min'
+            os: 'macOS-13' # Intel
```
devmotion:
```suggestion
            os: 'macOS-15-intel' # Intel
```

oxinabox:
> IMO such doctest failures should be avoided by using doctest filters or changing to more stable tests, since string representations may change even in non-breaking releases.

But the Manifest.toml doesn't lock to nonbreaking releases.
It locks to exact versions.

Should and _is_ are different things.
I agree ideally the doctests should use filters etc.
But that's not what the code as written does.

PR #248, opened 2026-03-29
Add compat for ChainRulesCore v2

JamesWrigley:
See https://github.com/JuliaDiff/ChainRulesCore.jl/pull/703. Tested manually by `dev`ing v2 and https://github.com/JuliaArrays/StaticArrays.jl/pull/1339.

The `to_vec()` inference test is failing on 1.12 so I disabled it.

### DIFF
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   test:
-    name: ${{ matrix.package.repo }}
+    name: ${{ matrix.package.name }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -15,12 +15,12 @@ jobs:
         julia-version: [1]
         os: [ubuntu-latest]
         package:
-          - {user: JuliaDiff, repo: ChainRulesTestUtils.jl}
-          - {user: JuliaDiff, repo: ChainRules.jl}
-          - {user: JuliaMath, repo: SpecialFunctions.jl}
-          - {user: invenia, repo: BlockDiagonals.jl}
-          - {user: invenia, repo: PDMatsExtras.jl}
-          - {user: PumasAI, repo: DataInterpolations.jl}
+          - {user: JuliaDiff, name: ChainRulesTestUtils}
+          - {user: JuliaDiff, name: ChainRules}
+          - {user: JuliaMath, name: SpecialFunctions}
+          - {user: JuliaArrays, name: BlockDiagonals}
+          - {user: invenia, name: PDMatsExtras}
+          - {user: SciML, name: DataInterpolations}
 
     steps:
       - uses: actions/checkout@v4
@@ -28,21 +28,22 @@ jobs:
         with:
           version: ${{ matrix.julia-version }}
           arch: x64
+      - uses: julia-actions/cache@v3
       - uses: julia-actions/julia-buildpkg@latest
       - name: Clone Downstream
         uses: actions/checkout@v4
         with:
-          repository: ${{ matrix.package.user }}/${{ matrix.package.repo }}
+          repository: ${{ matrix.package.user }}/${{ matrix.package.name }}.jl
           path: downstream
       - name: Load this and run the downstream tests
-        shell: julia --project=downstream {0}
+        shell: julia --project=@temp {0}
         run: |
           using Pkg
           try
             # force it to use this PR's version of the package
-            Pkg.develop(PackageSpec(path="."))  # ResolverError with main deps may be thrown here
+            Pkg.develop([PackageSpec(path="."), PackageSpec(path="downstream")])  # ResolverError with main deps may be thrown here
             Pkg.update()
-            Pkg.test()  # ResolverError with test deps may be thrown here (before test process starts)
+            Pkg.test("${{ matrix.package.name }}")  # ResolverError with test deps may be thrown here (before test process starts)
           catch err
             err isa Pkg.Resolve.ResolverError || rethrow()
             # If we can't resolve that means this is incompatible by SemVer and this is fine
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FiniteDifferences"
 uuid = "26cc04aa-876d-5657-8c51-4c34ba976000"
-version = "0.12.33"
+version = "0.12.34"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
@@ -12,7 +12,7 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 
 [compat]
-ChainRulesCore = "0.10.3, 1"
+ChainRulesCore = "0.10.3, 1, 2"
 LinearAlgebra = "<0.0.1, 1"
 Printf = "<0.0.1, 1"
 Random = "<0.0.1, 1"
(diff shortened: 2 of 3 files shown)


JamesWrigley:
The BlockDiagonals and SpecialFunctions tests should be fixed by these PRs:
- https://github.com/JuliaArrays/BlockDiagonals.jl/pull/151
- https://github.com/JuliaMath/SpecialFunctions.jl/pull/525

The amount of ChainRules failures is a bit alarming :grimacing: Nightly failures are the same kind of inference failure that's on 1.12. Is that worth a Julia issue?

test/to_vec.jl
```diff
@@ -190,7 +190,7 @@ end
             if T == Float64
                 test_to_vec((a=5, b=randn(10, 11), c=(5, 4, 3)); check_inferred = VERSION ≥ v"1.2")
             else
-                test_to_vec((a=3 + 2im, b=randn(T, 10, 11), c=(5+im, 2-im, 1+im)); check_inferred = VERSION ≥ v"1.2")
+                test_to_vec((a=3 + 2im, b=randn(T, 10, 11), c=(5+im, 2-im, 1+im)); check_inferred = VERSION ≥ v"1.2" && VERSION < v"1.12")
```
devmotion:
```suggestion
                test_to_vec((a=3 + 2im, b=randn(T, 10, 11), c=(5+im, 2-im, 1+im)); check_inferred = v"1.2" <= VERSION < v"1.12")
```

JamesWrigley: Fixed in 5411882.

.github/workflows/IntegrationTest.yml
```diff
@@ -7,20 +7,20 @@ on:
...
-          - {user: JuliaDiff, repo: ChainRulesTestUtils.jl}
-          - {user: JuliaDiff, repo: ChainRules.jl}
-          - {user: JuliaMath, repo: SpecialFunctions.jl}
-          - {user: invenia, repo: BlockDiagonals.jl}
-          - {user: invenia, repo: PDMatsExtras.jl}
-          - {user: PumasAI, repo: DataInterpolations.jl}
+          - {user: JuliaDiff, name: ChainRulesTestUtils}
+          - {user: JuliaDiff, name: ChainRules}
+          - {user: JuliaMath, name: SpecialFunctions}
+          - {user: invenia, name: BlockDiagonals}
```
devmotion:
```suggestion
          - {user: JuliaArrays, name: BlockDiagonals}
```

.github/workflows/IntegrationTest.yml
```diff
@@ -7,20 +7,20 @@ on:
...
+          - {user: invenia, name: PDMatsExtras}
+          - {user: PumasAI, name: DataInterpolations}
```
devmotion:
```suggestion
          - {user: SciML, name: DataInterpolations}
```

.github/workflows/IntegrationTest.yml
```diff
@@ -32,17 +32,18 @@ jobs:
...
         with:
-          repository: ${{ matrix.package.user }}/${{ matrix.package.repo }}
+          repository: ${{ matrix.package.user }}/${{ matrix.package.name }}.jl
           path: downstream
       - name: Load this and run the downstream tests
-        shell: julia --project=downstream {0}
+        shell: julia --project=@temp {0}
         run: |
           using Pkg
+          Pkg.develop(PackageSpec(path="."))
```
devmotion: No reason to do the `Pkg.develop` in two different places, they can be moved together (see below):


```suggestion
```

.github/workflows/IntegrationTest.yml
```diff
@@ -32,17 +32,18 @@ jobs:
...
           try
             # force it to use this PR's version of the package
-            Pkg.develop(PackageSpec(path="."))  # ResolverError with main deps may be thrown here
+            Pkg.develop(PackageSpec(path="downstream"))  # ResolverError with main deps may be thrown here
```
devmotion:
```suggestion
            Pkg.develop([PackageSpec(; path="."), PackageSpec(; path="downstream")])  # ResolverError with main deps may be thrown here
```

.github/workflows/IntegrationTest.yml
JamesWrigley: Fixed in aafc680.

.github/workflows/IntegrationTest.yml
JamesWrigley: Fixed in aafc680.

.github/workflows/IntegrationTest.yml
JamesWrigley: Removed in aafc680.

.github/workflows/IntegrationTest.yml
JamesWrigley: Fixed in aafc680.

JamesWrigley:
I also added the cache action in 2c6181c.

test/to_vec.jl
```diff
@@ -190,7 +190,7 @@ end
             if T == Float64
                 test_to_vec((a=5, b=randn(10, 11), c=(5, 4, 3)); check_inferred = VERSION ≥ v"1.2")
             else
-                test_to_vec((a=3 + 2im, b=randn(T, 10, 11), c=(5+im, 2-im, 1+im)); check_inferred = VERSION ≥ v"1.2")
+                test_to_vec((a=3 + 2im, b=randn(T, 10, 11), c=(5+im, 2-im, 1+im)); check_inferred = v"1.2" <= VERSION < v"1.12")
```
devmotion, line 193: Is it clear why this is broken?

Project.toml
devmotion, line 1: This (or rather the integration test fixes) should be done in a separate PR.

test/to_vec.jl
JamesWrigley: I ran it through Claude and it said that it was something to do with struggling to infer the return of `reduce(vcat)` with heterogeneous tuples. But I'm not familiar enough with the code to say if that's correct :shrug:

Project.toml
JamesWrigley: Why? :thinking: They're in a separate commit.

devmotion: Because the CI changes don't deserve a new release, and ChainRulesCore@2 is not even released and can't be tested. The CI changes are useful right away.

JamesWrigley: That's why I tested it manually by `dev`ing v2 :) How then do you want to handle https://github.com/JuliaDiff/ChainRulesCore.jl/pull/703? Temporarily dev FiniteDifferences/StaticArrays/ChainRulesTestUtils in the ChainRulesCore tests? Otherwise CI won't pass.

devmotion: It's fine to prepare PRs but they can't be merged until a new release is available. CI fixes should not wait for it.

Generally, to test unreleased versions/branches in CI, you should use the `sources` feature of Pkg.

JamesWrigley: Ok :+1: If it's alright with you I'll keep everything in these PRs for now including the version bumps, otherwise I'll have to do another round of PRs later. I'll tweak the ChainRulesCore CI to use these branches for now.

JamesWrigley: Also `sources` was only added in 1.11 so 1.10 CI will fail. Is that ok?

devmotion: I still don't see a reason for adding unrelated changes that master would benefit from right now to this PR with unclear time horizon and without proper CI test of the remaining changes.

test/to_vec.jl
devmotion: The issue should either be fixed or marked as broken. Just disabling failing tests is not a proper solution.

devmotion:
I think there are two main outstanding issues here:
1. Move the non-ChainRulesCore related changes to a separate PR
2. Consider making a ChainRulesCore extension?

JamesWrigley:
> 1. Move the non-ChainRulesCore related changes to a separate PR

Why exactly? There's a lot of PRs I'm having to keep track of and it's extra work for me.

> 2. Consider making a ChainRulesCore extension?

I would suggest doing that in another PR.

oxinabox (dismissed):

